PR #53, opened 2023-03-10
optimize set_coefficient PolyOverZ

Marvin-Beckmann:
This PR removes one drop operation of the input value, if the coefficient itself is set before. If the entry itself is not yet set, we continue to use fmpz_poly_set_coeff for now. fmpz_poly_zero_coeff can be used to set entries manually to 0, but all leading entries with zero are automatically removed. Any further suggestions are welcome to optimize the other case as well.

One Drop is called less because: Before: fmpz_poly_set_coeff drops the previous value and sets the coefficient to the new value. Additionally the old value remains and has to be dropped as well. If we swap the allocated memory, only the value after the swap has to be dropped -> one drop less.


Marvin-Beckmann:
Approach from before not possible with function signature, hence we keep the current version for now

PR #105, opened 2023-03-22
Resolve TODO in From_str Modulus

Marvin-Beckmann:
This PR resolves a TODO in from_str for Modulus that remained since Z did not had to_string() nad from_str() at the time of implementation.
Additionally this PR introduces the constant ZERO for Z, which is especially useful for comparisons and increases readability.
This can be seen as separate from the default value, since we expect that the constant ZERO will not change, whereas the Default could potentially be changed.

- [x] resolve TODO
- [x] remove leftover error InvalidStringToModulus
- [x] add ZERO constant

### DIFF
--- a/src/integer_mod_q/modulus/from.rs
+++ b/src/integer_mod_q/modulus/from.rs
@@ -102,7 +102,7 @@ impl FromStr for Modulus {
 /// - Returns a [`MathError`] of type [`InvalidIntToModulus`](MathError::InvalidIntToModulus)
 /// if the provided value is not greater than `0`.
 fn ctx_init(n: &Z) -> Result<fmpz_mod_ctx, MathError> {
-    if n <= &Z::from(0) {
+    if n <= &Z::ZERO {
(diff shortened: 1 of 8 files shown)


src/integer_mod_q/modulus/from.rs
```diff
@@ -77,56 +74,40 @@ impl FromStr for Modulus {
     /// # Errors and Failures
     ///
     /// - Returns a [`MathError`] of type
-    /// [`InvalidStringToModulusInput`](MathError::InvalidStringToModulusInput) if the provided string was not
-    /// formatted correctly, e.g., not a number or not greater `0`.
-    /// - Returns a [`MathError`] of type [`InvalidIntToModulus`](MathError::InvalidIntToModulus)
+    /// [`InvalidStringToZ`](MathError::InvalidStringToZInput) if the
```
Marcel583, line 77:
```suggestion
    /// [`InvalidStringToZInput`](MathError::InvalidStringToZInput) if the
```

Marcel583 (approved):

PR #121, opened 2023-03-28
add exponential function using taylor polynomial

Marvin-Beckmann:
This PR introduces the ability to approximate the value of the exponential function for a given value. This works well for small values, but becomes less precise for large valeus. I will proceed by looking at different options, but this one allows to approximate the value of the exponential function easily using the Taylor series.

- [x] define taylor series
- [x] define exponential function using taylor series for Z and Q

### DIFF
--- a/src/integer/z/exp.rs
+++ b/src/integer/z/exp.rs
@@ -6,7 +6,7 @@
 // the terms of the Mozilla Public License Version 2.0 as published by the
 // Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
 
-//! Implementations to call the exponential function on a [`Z`].
+//! Implementations to call the exponential function on a [`Z`] integer.
 
 use super::Z;
 use crate::{
--- a/src/integer/z/exp.rs
+++ b/src/integer/z/exp.rs
@@ -1,0 +1,62 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Implementations to call the exponential function on a [`Z`].
+
+use super::Z;
+use crate::{
+    rational::{PolyOverQ, Q},
+    traits::Evaluate,
+};
+
+impl Z {
+    /// Computes e^self using taylor series approximation of the exponential function.
+    ///
+    /// Parameters:
+    /// - `length_taylor_polynomial`: the length of the taylor series
+    /// approximation of the exponential function
+    ///
+    /// Returns e^self.
+    ///
+    /// # Example
+    /// ```
+    /// use math::integer::Z;
+    ///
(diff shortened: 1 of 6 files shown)


src/integer/z/exp.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Implementations to call the exponential function on a [`Z`].
```
Marcel583, line 9:
```suggestion
//! Implementations to call the exponential function on a [`Z`] integer.
```

src/rational/poly_over_q/exp.rs
```diff
@@ -0,0 +1,72 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Implementations to approximate the exponential function using a [`PolyOverQ`].
```
Marcel583, line 9:
```suggestion
//! Implementations to approximate the exponential function using a [`PolyOverQ`] polynomial.
```

src/rational/poly_over_q/exp.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// - `length`: the length of the taylor series approximation of the exponential function
+    ///
+    /// Returns the Taylor approximation of the exponential function.
+    ///
+    /// # Example
+    /// ```compile_fail
+    /// use math::rational::PolyOverQ;
+    ///
+    /// // sum_{k=0}^{length-1} x^k/k!
+    /// let taylor_approximation_exponential_function = PolyOverQ::exp_function(1000);
```
Marcel583, line 28:
```suggestion
    /// let taylor_approximation_exponential_function = PolyOverQ::exp_function_taylor(1000);
```

src/rational/poly_over_q/exp.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// ```
+    pub(crate) fn exp_function_taylor(length: u32) -> Self {
```
Marcel583, line 30: Maybe not just u32, but something that can be turned into u32

src/rational/q/exp.rs
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// use math::rational::Q;
+    /// use std::str::FromStr;
+    ///
+    /// // sum_{k=0}^999 (17/3)^k/k!
+    /// let evaluation = Q::from_str("17/3").unwrap().exp_taylor(1000);
+    /// ```
+    pub fn exp_taylor(&self, length_taylor_polynomial: u32) -> Self {
```
Marcel583, line 31: s.o.

Marcel583 (approved):

PR #124, opened 2023-03-28
From int for `Q`

MoogSven:
Implement `Q::try_from` just for two borrowed integers.

The current implementation is not efficient and also not supporting
owned parameters.
Since the `Q` from integers functionality is needed, I decided to push this imperfect solution. 

A medium/larger rewrite might be required to accept all combinations of owned/borrowed, data types, and also avoid unnecessary clones.
The interface of the Trait should just accept more combinations after the rewrite. Therefore, the current `Q::try_from` trait implementation can be used without causing problems in the future.

### DIFF
--- a/src/rational/q/from.rs
+++ b/src/rational/q/from.rs
@@ -111,8 +111,11 @@ impl FromStr for Q {
 }
 
 impl Q {
-    /// Create a [`Q`] from two values that can be converted to [`Z`].
-    /// For example, [`i64`] and [`Z`].
+    /// Create a [`Q`] from two references that can be converted to [`Z`].
+    /// For example, [`&Z`].
+    ///
+    /// Warning: The interface of this function will likely change in the future
+    /// or it will entirely be removed. Therefore use [`Q::try_from`] instead.
     ///
     /// Parameters:
     /// - `numerator` of the new [`Q`].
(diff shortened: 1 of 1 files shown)


src/rational/q/from.rs
```diff
@@ -110,6 +110,100 @@ impl FromStr for Q {
...
+    /// let num = Z::from(100);
+    ///
+    /// let a = Q::try_from_int_int_borrow(&num, &i64::MAX).unwrap();
+    /// let b = Q::try_from_int_int_borrow(&num, &i64::MAX).unwrap();
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`DivisionByZeroError`](MathError::DivisionByZeroError)
+    /// if the denominator is zero.
+    fn try_from_int_int_borrow(
```
jnsiemer, line 140:
```suggestion
    fn try_from_int_int(
```
As you do not implement this function for owned types, just leave the `_borrow` at the end out.

src/rational/q/from.rs
```diff
@@ -110,6 +110,100 @@ impl FromStr for Q {
...
+
+        // TODO: this is not as efficient as possible when passing values that
+        // internally include a [`fmpz`], e.g. [`Z`] or [`Zq`].
+        // In those cases it would be faster to use `fmpq_set_fmpz_frac`.
+        // The best way would probably be to use `Into<fmpz>` and do the
+        // performance improvements there. However, this takes longer to implement and this
+        // functionality is now required for others to make progress.
+
+        // TODO: add [`num::Zero`] trait and use is_zero
+        if denominator == Z::from(0) {
```
jnsiemer, line 155:
```suggestion
        if denominator == Z::ZERO {
```

src/rational/q/from.rs
```diff
@@ -249,3 +343,105 @@ mod tests_from_str {
...
+    #[test]
+    fn test_types_borrowed() {
+        let numerator = 10;
+        let denominator = 15;
+
+        let q_1 = Q::try_from((&(numerator as u8), &(denominator as i8))).unwrap();
+        let q_2 = Q::try_from((&(numerator as u16), &(denominator as i16))).unwrap();
+        let q_3 = Q::try_from((&(numerator as u32), &(denominator as i32))).unwrap();
+        let q_4 = Q::try_from((&(numerator as u64), &(denominator as i64))).unwrap();
+        let q_5 = Q::try_from((&Z::from(numerator), &Z::from(denominator))).unwrap();
```
jnsiemer, line 363: Please also add tests where the numerator is a signed integer and the denominator is unsigned. It does not have to match all combinations, but at least all combinations (sigend, unsigned), (unsigned, unsigned), (signed, signed) should be added with one example

src/rational/q/from.rs
```diff
@@ -249,3 +343,105 @@ mod tests_from_str {
...
+    #[test]
+    fn numerator_and_denominator_negative_large() {
+        let numerator = i64::MAX;
+        let denominator = i64::MAX - 1;
+
+        let q_1 = Q::try_from((&numerator, &denominator)).unwrap();
+        let q_2 = Q::try_from((&-numerator, &-denominator)).unwrap();
+
+        assert_eq!(q_1, q_2);
+    }
```
jnsiemer, line 446: These tests check more than you describe. Please adapt the comment (only needed in the first case) and function name accordingly

jnsiemer (changes requested):
Great work, just minor details that should be worked out a bit more

src/rational/q/from.rs
MoogSven: Added

src/rational/q/from.rs
```diff
@@ -110,6 +110,100 @@ impl FromStr for Q {
...
+    /// ```rust
+    /// use math::rational::Q;
+    /// use math::integer::Z;
+    ///
+    /// let a = Q::try_from((&42, &2)).unwrap();
+    /// let b = Q::try_from((&Z::from(21), &Z::from(1))).unwrap();
+    /// assert_eq!(a,b);
+    ///
+    /// let c = Q::try_from((&10,&0));
+    /// assert!(c.is_err());
```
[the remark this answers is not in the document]
MoogSven: Added

jnsiemer (approved):

PR #125, opened 2023-03-28
`PartialEq` for MatQ

MoogSven:
Add `PartialEq` for `MatQ`

### DIFF
--- a/src/rational/mat_q/cmp.rs
+++ b/src/rational/mat_q/cmp.rs
@@ -1,0 +1,109 @@
+// Copyright © 2023 Sven Moog, Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains implementations for comparison of [`MatQ`].
+
+use super::MatQ;
+use flint_sys::fmpq_mat::fmpq_mat_equal;
+
+impl PartialEq for MatQ {
+    /// Checks if two [`MatQ`] instances are equal. Used by the `==` and `!=` operators.
+    ///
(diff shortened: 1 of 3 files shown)


src/rational/mat_q/cmp.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright © 2023 Sven Moog, Niklas Siemer
```
jnsiemer, line 1:
```suggestion
// Copyright © 2023 Sven Moog
```

src/rational/mat_q/cmp.rs
```diff
@@ -0,0 +1,109 @@
...
+mod test_partial_eq {
+
+    use super::MatQ;
+    use crate::rational::Q;
+    use std::str::FromStr;
+
+    /// Ensures that different instantiations do not break the equality between matrices
+    #[test]
+    fn equality_between_instantiations() {
+        let a = MatQ::from_str("[[0,1],[0,0]]").unwrap();
```
jnsiemer, line 63: Please add a denominator in at least one of the tests as you instantiate `MatQ` and this is its most important feature.
Especially testworthy is something like 1/2 ==2/4. Does the equality hold in the case of matrices?

src/rational/mat_q/cmp.rs
```diff
@@ -0,0 +1,109 @@
...
+        ))
+        .unwrap();
+
+        assert_eq!(&a, &b);
+    }
+
+    /// Checks that different unequal matrices are unequal
+    #[test]
+    fn not_equal() {
+        let a = MatQ::from_str(&format!("[[{},{}],[-10, 10]]", i64::MIN, i64::MAX)).unwrap();
```
jnsiemer, line 92: Same here, please add one with a denominator

jnsiemer (changes requested):
Nice work, just minor additions required to test specific rational cases - and not just integer ones :)

jnsiemer (approved):

PR #126, opened 2023-03-29
add evaluation of MatPolyOverZ

Marvin-Beckmann:
This PR introduces the ability to evaluate a MatPolyOverZ entrywise with every reasonable type.

### DIFF
--- a/src/integer/mat_poly_over_z/evaluate.rs
+++ b/src/integer/mat_poly_over_z/evaluate.rs
@@ -95,6 +95,17 @@ mod test_evaluate {
         assert_eq!(MatZ::from_str(&res_cmp_str).unwrap(), res)
     }
 
+    /// tests whether evaluation of a large entry in the matrix with a large value works
+    #[test]
+    fn eval_large_with_large() {
+        let poly_str = format!("[[2  {} -1, 1  1],[1  {}, 2  0 1]]", u64::MAX, u64::MAX);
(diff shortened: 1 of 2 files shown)


src/integer/mat_poly_over_z/evaluate.rs
```diff
@@ -0,0 +1,135 @@
...
+    #[test]
+    fn eval_z_large() {
+        let poly_str = "[[1  17],[2  6 2]]";
+        let poly = MatPolyOverZ::from_str(poly_str).unwrap();
+
+        let res = poly.evaluate(&Z::from_str(&"1".repeat(65)).unwrap());
+
+        let res_cmp_str = format!("[[17],[{}8]]", "2".repeat(64));
+        assert_eq!(MatZ::from_str(&res_cmp_str).unwrap(), res)
+    }
```
jnsiemer, line 96: It would be great if you added a test, where the poly_str includes a large value and is evaluated with either a small and/or a large value. This case should probably not make any problems, but it eventually could.

jnsiemer (commented):
Great work. There's just one addition in the test section that I would love to see.

src/integer/mat_poly_over_z/evaluate.rs
Marvin-Beckmann: added it, please have another look over it

jnsiemer (approved):

PR #129, opened 2023-03-29
refactor Matrices

Marcel583:


src/integer/mat_z/get.rs
```diff
@@ -142,18 +145,13 @@ mod test_get_entry {
...
+            .set_entry(1, 1, Z::from_str(&format!("-{}", u64::MAX)).unwrap())
             .unwrap();
 
         let entry = matrix.get_entry(1, 1).unwrap();
 
-        let mut test_entry = "-".to_string();
-        test_entry.push_str(&"1".repeat(65));
-
-        assert_eq!(Z::from_str(&test_entry).unwrap(), entry);
+        assert_eq!(Z::from_str(&format!("-{}", u64::MAX)).unwrap(), entry);
```
MoogSven, line 154:
```suggestion
        assert_eq!(Z::from(i64::MIN)).unwrap(), entry);
```

src/rational/mat_q/get.rs
```diff
@@ -15,6 +15,38 @@ use crate::{error::MathError, rational::Q};
 use flint_sys::{fmpq::fmpq_set, fmpq_mat::fmpq_mat_entry};
 use std::fmt::Display;
 
+impl GetNumRows for MatQ {
+    /// Returns the number of rows of the matrix as a [`i64`].
```
MoogSven, line 19: Would it be better to return it as a u64? (I am not sure)

Advantages:
- It represents the data better
- Using unsigned integers later might allow to avoid a check that a value is positive when accessing an index.

Disadvantages:
- not what flint uses internally

MoogSven (changes requested):
Wasn't able to finish the review yet, but here are some comments so far:
So far: looks good

Before, `set_entry` was (sometimes) divided into an implementation that takes ownership and one that just takes a reference. This allows to optimize the code if ownership is passed. This could be added by implementing the `set_entry` trait for the reference and ownership case separably and not using `implement_for_owned!`.

Also: I think that a Matrix trait that combines the other traits is possible. (https://github.com/qfall/math/commit/640025ee111334f62d62c7ab0cb4db943d54121f)

src/integer/mat_z/get.rs
Marcel583: s.o.

src/integer/mat_poly_over_z/set.rs
```diff
@@ -28,53 +28,17 @@ impl MatPolyOverZ {
...
-        let (row_i64, column_i64) = evaluate_coordinates(self, row, column)?;
-
-        // swapping the content of the entry with the given value since ownership
-        // of the input is provided.
-        unsafe {
-            let entry = fmpz_poly_mat_entry(&self.matrix, row_i64, column_i64);
-            fmpz_poly_swap(entry, &mut value.poly)
-        }
-        Ok(())
-    }
```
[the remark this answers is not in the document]
Marcel583: It is not an optimization, the is in both cases cloned

src/rational/mat_q/get.rs
Marcel583: We had this decision in the challenge, flint uses i64 and we didn't want to convert all the time (short version)

src/integer/mat_z/get.rs
```diff
@@ -142,18 +145,13 @@ mod test_get_entry {
...
     }
```
MoogSven, line 155: Using a variable to store the string like `value_str` makes it more clear that the strings are the same. 

Also not fitting into i64 does not really change the behavior compared to the previous test case. 
However, not fitting into 64 bit could in theory. I personally would also be fine without this test case, since the different behavior when flint uses pointers is already covered by the previous test.
```rust
    /// Ensure that getting entries works with large negative numbers (not fitting into 64 bit).
    #[test]
    fn big_negative() {
        let mut matrix = MatZ::new(5, 10).unwrap();
        let value_str = &format!("-{}", u64::MAX); // value is not fitting into 64 bit
        matrix
            .set_entry(1, 1, Z::from_str(value_str).unwrap())
            .unwrap();

        let entry = matrix.get_entry(1, 1).unwrap();

        assert_eq!(Z::from_str(value_str).unwrap(), entry);
    } 
```
(could not make this as a code suggestion since the changed lines include deletions)

MoogSven (changes requested):

src/integer/mat_poly_over_z/set.rs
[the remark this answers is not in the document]
MoogSven: A way to add this optimization is to use the removed code to implement the `SetEntry<Z>` trait and not use the `implement_for_owned` macro.

PR #130, opened 2023-03-30
Vector Functions for MatZ

jnsiemer:
This PR implements...
- [X] is_vector, is_row_vector, is_column_vector + has_single_entry
- [ ] norm_sqrd_eucl, norm_infty (subsequent PR)
- [ ] dot_product (subsequent PR)

including tests for `MatZ`.

Furthermore, I added a function `collect_entries` that efficiently collects all matrix entries without cloning any objects (incl. test).

### DIFF
--- a/src/integer/mat_z/vector/is_vector.rs
+++ b/src/integer/mat_z/vector/is_vector.rs
@@ -1,0 +1,163 @@
+// Copyright © 2023 Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module includes all functionality to check
+//! whether a matrix represents a vector, i.e. has only one row
+//! or one column.
+//! These methods should be used to ensure that vector functions
+//! can only be called on suitably formed vector/matrices.
+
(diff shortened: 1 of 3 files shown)


src/integer/mat_z/vector/is_vector.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// use math::integer::MatZ;
+    /// use std::str::FromStr;
+    ///
+    /// let vec = MatZ::from_str("[[1]]").unwrap();
+    ///
+    /// assert!(vec.has_single_entry());
+    /// ```
+    pub fn has_single_entry(&self) -> bool {
+        self.is_column_vector() && self.is_row_vector()
+    }
```
Marvin-Beckmann, line 84: I do not quite get the pupose of this function, but if yout think that we need it, feel free to keep it

Marvin-Beckmann (approved):
Looks good. Please add an example for the vectors to the examples of the MatZ struct. This can also be included in a following PR

src/integer/mat_z/vector/is_vector.rs
jnsiemer: This function could become interesting, e.g. we want to automatically unpack single entry matrices for arithmetic operations. But yes, currently, we do not have any use cases.

jnsiemer:
> Looks good. Please add an example for the vectors to the examples of the MatZ struct. This can also be included in a following PR

Good idea. I will do that in a following PR.

PR #134, opened 2023-03-30
Serialize MatZq MatQ

Marvin-Beckmann:
This PR adds the serialize and deserialize method to MatZq and MatQ.

### DIFF
--- a/src/rational/mat_q.rs
+++ b/src/rational/mat_q.rs
@@ -15,6 +15,7 @@ mod cmp;
 mod from;
 mod get;
 mod ownership;
+mod serialize;
 mod set;
 mod to_string;
 
--- a/src/rational/mat_q/serialize.rs
+++ b/src/rational/mat_q/serialize.rs
@@ -1,0 +1,142 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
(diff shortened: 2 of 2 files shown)


src/integer_mod_q/mat_zq/serialize.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// tests whether the serialization of a positive large [`MatZq`] works.
+    #[test]
+    fn serialize_output_positive_large() {
+        let mat_poly_str = format!("[[3, 17, {}, 1, 2, 13, 5]] mod {}", u64::MAX - 1, u64::MAX);
+        let mat_poly_z = MatZq::from_str(&mat_poly_str).unwrap();
+        let cmp_string = format!("{{\"matrix\":\"{}\"}}", mat_poly_str);
+
+        assert_eq!(cmp_string, serde_json::to_string(&mat_poly_z).unwrap())
+    }
```
jnsiemer, line 60: Please add a test for negative large values for this case as you did in all other cases, too. (Just to stay consistent along these functions.

jnsiemer (approved):
Great work! Please add that one test just to have consistency along the (de-)serialize functions also along their tests.

src/integer_mod_q/mat_zq/serialize.rs
Marvin-Beckmann: ahh, saw it too late, will make a PR for it...

PR #137, opened 2023-03-31
MatQ is_vector

jnsiemer:
This PR implements 
- [X] `is_vector` including tests for `MatQ`. 
- [X] `collect_entries` for `MatQ`, which efficiently collects all `fmpq` entries from the matrix without cloning elements.
- [X] doc tests for `MatQ`

### DIFF
--- a/src/rational/mat_q/get.rs
+++ b/src/rational/mat_q/get.rs
@@ -96,6 +96,10 @@ impl MatQ {
     #[allow(dead_code)]
     /// Efficiently collects all [`fmpz`]s in a [`MatQ`] without cloning them.
     ///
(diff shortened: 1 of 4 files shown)


src/rational/mat_q/get.rs
```diff
@@ -89,6 +92,38 @@ impl GetNumColumns for MatQ {
     }
 }
 
+impl MatQ {
+    #[allow(dead_code)]
+    /// Efficiently collects all [`fmpz`]s in a [`MatQ`] without cloning them.
```
Marvin-Beckmann, line 97:
```suggestion
    /// Efficiently collects all [`fmpq`]s in a [`MatQ`] without cloning them.
```

PR #139, opened 2023-03-31
MatPolyOverZ is_vector

jnsiemer:
This PR implements 
- [X] `is_vector` including tests for `MatPolyOverZ`. 
- [X] `collect_entries` for `MatPolyOverZ`, which efficiently collects all `fmpz_poly_struct` entries from the matrix without cloning elements.
- [X] doc tests for `MatPolyOverZ`

At first, I did not want to implement this. But as the `vector` part probably get's added to the `Matrix` trait, it should be implemented. Additionally, with a norm defined for polynomials (see [here](https://mathworld.wolfram.com/PolynomialNorm.html)), it seems realistic, that we will need this in the future.

### DIFF
--- a/src/integer/mat_poly_over_z/get.rs
+++ b/src/integer/mat_poly_over_z/get.rs
@@ -285,10 +285,9 @@ mod test_collect_entries {
         let entries_2 = mat_2.collect_entries();
 
         assert_eq!(entries_1.len(), 6);
-        // 4611686018427387904 = 2^62, i.e. value is stored on stack
         assert_eq!(unsafe { *entries_1[0].coeffs }.0, 1);
-        assert!(unsafe { *entries_1[2].coeffs }.0 >= 4611686018427387904);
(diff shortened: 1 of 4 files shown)


src/integer/mat_poly_over_z/get.rs
```diff
@@ -230,3 +265,37 @@ mod test_get_num {
...
+            i64::MIN
+        ))
+        .unwrap();
+        let mat_2 = MatPolyOverZ::from_str("[[1  -1, 2  1 2]]").unwrap();
+
+        let entries_1 = mat_1.collect_entries();
+        let entries_2 = mat_2.collect_entries();
+
+        assert_eq!(entries_1.len(), 6);
+        // 4611686018427387904 = 2^62, i.e. value is stored on stack
```
Marvin-Beckmann, line 288:
```suggestion
```

Marvin-Beckmann (approved):
good work

PR #116, opened 2023-03-27
arith between types

3pmile:

PR #140, opened 2023-04-02
Arith z standard types

3pmile:


3pmile:
rebased via  cerrypicking

3pmile:
old: https://github.com/qfall/math/pull/116,
resolved all reviews, $output_type is correct, reason: 

 the documentation should be shown at the output type (the suggestion would link Z+Q to Z and Q+Z to Q, which is unintuitive), maybe an update of the documentation with new examples is needed, todo for Q and Zq

src/integer/z/arithmetic/add.rs
```diff
@@ -46,8 +47,102 @@ impl Add for &Z {
...
+mod test_add_between_types {
+
+    use crate::integer::Z;
+    use std::str::FromStr;
+
+    /// testing addition between different types
+    #[test]
+    #[allow(clippy::op_ref)]
+    fn add() {
+        let a: Z = Z::from(42);
```
Marvin-Beckmann, line 64: can you do it for the others as well and not just this one?

Marvin-Beckmann (approved):

PR #160, opened 2023-04-07
Tensor product for MatZ

Marvin-Beckmann:
This PR adds the ability to compute the tensor product of two matrices: $A \otimes B$. For reference regarding correct implementation, one can look [here](https://en.wikipedia.org/wiki/Tensor_product).

- [x] Tensor trait
- [x] MatZ Tensor trait implementation

### DIFF
--- a/src/integer/mat_z/tensor.rs
+++ b/src/integer/mat_z/tensor.rs
@@ -1,0 +1,188 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains the implementation of the `tensor` product.
+
+use super::MatZ;
+use crate::traits::{GetNumColumns, GetNumRows, Tensor};
+use flint_sys::{
+    fmpz::{fmpz_is_zero, fmpz_mul},
+    fmpz_mat::fmpz_mat_entry,
+};
+
+impl Tensor for MatZ {
+    /// Computes the tensor product of `self` with `other`
+    ///
+    /// Parameters:
+    /// - `other`: the value with which the tensor product is computed.
+    ///
+    /// Returns the tensor product of `self` with `other`.
+    ///
+    /// # Example
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    /// use qfall_math::traits::Tensor;
+    /// use std::str::FromStr;
+    ///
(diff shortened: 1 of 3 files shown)


src/integer/mat_z/tensor.rs
```diff
@@ -0,0 +1,188 @@
...
+                                        i * rows_other + i_other,
+                                        j * columns_other + j_other,
+                                    ),
+                                    entry,
+                                    fmpz_mat_entry(&other.matrix, i_other, j_other),
+                                )
+                            }
+                        }
+                    }
+                }
```
jnsiemer, line 70: please outsource this part:
- the nicest way would be via two efficient functions in the fmpz_helpers.rs file in src/integer/z/: one for concatenation of matrices and one for scalar multiplication (probably not as efficient as your current implementation without having several trade offs)
- just one function specifically for this.

It's just too many for-loops in each other/ tabs in one line.

jnsiemer (approved):
Great work, please consider my comment before merging

src/integer/mat_z/tensor.rs
Marvin-Beckmann: I prefer the second version, since it is probably more efficient. While programming I considered your first idea, but due to efficiency scraped that idea

Marvin-Beckmann: I will add a subroutine for the latter

src/integer/mat_z/tensor.rs
```diff
@@ -0,0 +1,214 @@
...
+/// Sets the entries
+/// `[i*rows_other, j*columns_other]` up till `[i*(row_other +1), j*(columns_other + 1)]`
+/// according to the definition of the tensor product
+///
+/// Parameters:
+/// - `out`: the matrix in which the result is entered
+/// (**Warning**: even though `out` is not mutable, the subpart of the matrix is still
+/// set since the content behind the pointers changed and not the pointers themselves)
+/// - `i`: defines the row from which `entry` is taken
+/// - `j`: defines the column from which `entry` is taken
```
jnsiemer, line 73: Please describe these a little more precise, e.g. `defines the leftmost row of the set window`

src/integer/mat_z/tensor.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// use qfall_math::traits::Tensor;
+    /// use std::str::FromStr;
+    ///
+    /// let mat_1 = MatZ::from_str("[[1, 0, 0],[0, 1, 0],[0, 0, 1]]").unwrap();
+    /// let mat_2 = MatZ::from_str("[[1, 2, 1],[3, 4, 1]]").unwrap();
+    ///
+    /// let mat_3 = mat_1.tensor(&mat_2);
+    /// ```
+    fn tensor(&self, other: &Self) -> Self {
+        // we can unwrap since we know that the dimensions are positive
```
jnsiemer, line 38:
```suggestion
```
I don't think this comment is necessary anymore

PR #165, opened 2023-04-08
MatZq additional attribute `modulus`

jnsiemer:
This PR adds an additional attribute `modulus` to the `MatZq` struct. It allows for a more convenient way to `get_mod`, use this modulus efficiently, and not care about memory management issues, which arised from dropped Z/ Modulus values before.

### DIFF
--- a/src/integer_mod_q/mat_zq.rs
+++ b/src/integer_mod_q/mat_zq.rs
@@ -9,6 +9,7 @@
 //! [`MatZq`] is a type of matrix with integer entries of arbitrary length modulo `q`.
 //! This implementation uses the [FLINT](https://flintlib.org/) library.
 
+use crate::integer_mod_q::Modulus;
 use flint_sys::fmpz_mod_mat::fmpz_mod_mat_struct;
 
 mod cmp;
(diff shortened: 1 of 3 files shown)


src/integer_mod_q/mat_zq.rs
```diff
@@ -67,4 +68,12 @@ mod vector;
...
 pub struct MatZq {
     pub(crate) matrix: fmpz_mod_mat_struct,
+
+    // Since `get_mod` is needed for almost any action with `MatZq` instances
+    // a separate storage of the modulus object guarantees memory safety and
+    // is due to a reference counter in the `Modulus` object memory efficient.
+    // The modulus of a `MatZq` is not able to be modified afterwards. Hence, we
+    // do not need to care about conformity of the modulus stored in the `matrix`
+    // attribute and `modulus` attribute, if they are both set from the same value.
+    pub(crate) modulus: Modulus,
```
MoogSven, line 78:
```suggestion
    modulus: Modulus,
```
Making the modulus private forces us to use get_mod. It also prevents setting the modulus to a different value.

MoogSven (approved):
Looks good, just small suggestions

PR #166, opened 2023-04-08
Z `distance`

jnsiemer:
This PR refactors the implementation of `distance` for `Z`.
It implements a new trait `Distance` s.t. it can also be used for owned, borrowed, and u8, u16,... types.

This PR prepares an implementation of the `distance` trait for all numeric types.

### DIFF
--- a/src/integer/z/distance.rs
+++ b/src/integer/z/distance.rs
@@ -10,6 +10,7 @@
 
 use super::Z;
 use crate::{
+    integer_mod_q::{Modulus, Zq},
     macros::for_others::{implement_for_others, implement_for_owned},
     traits::Distance,
 };
@@ -44,12 +45,12 @@ impl Distance<&Z> for Z {
     }
 }
 
-implement_for_others!(Z, Z, Distance for u8 u16 u32 u64 i8 i16 i32 i64);
 implement_for_owned!(Z, Z, Distance);
+implement_for_others!(Z, Z, Distance for u8 u16 u32 u64 i8 i16 i32 i64 Modulus Zq);
 
 #[cfg(test)]
 mod test_distance {
-    use super::{Distance, Z};
+    use super::{Distance, Modulus, Zq, Z};
 
     /// Checks if distance is correctly output for small [`Z`] values
     /// and whether distance(a,b) == distance(b,a), distance(a,a) == 0
@@ -89,6 +90,8 @@ mod test_distance {
     #[test]
     fn other_types() {
         let a = Z::ZERO;
+        let modulus = Modulus::try_from_z(&Z::ONE).unwrap();
+        let zq = Zq::try_from_int_int(15, 19).unwrap();
 
         let u_0 = a.distance(0_u8);
         let u_1 = a.distance(15_u16);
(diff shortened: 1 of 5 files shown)


src/integer/z/distance.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+implement_for_owned!(Z, Z, Distance);
+implement_for_others!(Z, Z, Distance for u8 u16 u32 u64 i8 i16 i32 i64 Modulus Zq);
+
+#[cfg(test)]
+mod test_distance {
+    use super::{Distance, Modulus, Zq, Z};
+
+    /// Checks if distance is correctly output for small [`Z`] values
```
Marvin-Beckmann, line 55:
```suggestion
    /// Checks if distance is correctly returned for small [`Z`] values
```

src/integer/z/distance.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(Z::ONE, a.distance(&zero));
+        assert_eq!(Z::ONE, zero.distance(&a));
+        assert_eq!(Z::from(16), a.distance(&b));
+        assert_eq!(Z::from(16), b.distance(&a));
+        assert_eq!(Z::from(15), b.distance(&zero));
+        assert_eq!(Z::from(15), zero.distance(&b));
+        assert_eq!(Z::ZERO, b.distance(&b))
+    }
+
+    /// Checks if distance is correctly output for large [`Z`] values
```
Marvin-Beckmann, line 72:
```suggestion
    /// Checks if distance is correctly returned for large [`Z`] values
```

src/macros/for_others.rs
```diff
@@ -109,6 +109,21 @@ macro_rules! implement_for_others {
             }
         })*
     };
+
```
Marvin-Beckmann, line 112: Doc comments for macros is missing together with the corresponding examples

src/traits.rs
```diff
@@ -126,3 +126,14 @@ pub trait Concatenate {
...
+
+pub trait Distance<T> {
+    /// Computes the absolute distance between two values.
+    ///
+    /// Parameters:
+    /// - `other`: specifies the value whose distance is calculated to `self`
+    ///
+    /// Returns the absolute difference, i.e. distance between the two given values
+    /// as a new instance.
+    fn distance(&self, other: T) -> Self;
```
Marvin-Beckmann, line 138: I am not quite shure if we always want self as a return value. I think of Zq where we might want Z as a distance measurement. Please consider changing this trait implementation to specifiy an output type rather than directly returning Self

Marvin-Beckmann (changes requested):
Great, just some minor tweaks

src/integer/z/distance.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// # assert_eq!(Z::from(11), distance_1);
+    /// ```
+    fn distance(&self, other: &Z) -> Self::Output {
+        let difference = other - self;
+        difference.abs()
+    }
+}
+
+implement_for_owned!(Z, Z, Distance);
+implement_for_others!(Z, Z, Distance for u8 u16 u32 u64 i8 i16 i32 i64 Modulus Zq);
```
Marvin-Beckmann, line 51: Do you want Zq to be implemented by this macro? I personally would not be quite shur how the distance is measured in theis case, hence I think that the simple Macro implementation seems off here. I would remove it, or implement it separately with a dedicated description on the expected output
```suggestion
implement_for_others!(Z, Z, Distance for u8 u16 u32 u64 i8 i16 i32 i64 Modulus);
```

Marvin-Beckmann (approved):

PR #168, opened 2023-04-09
Z `gcd` + `xgcd`

jnsiemer:
This PR implements...
- [X] the greatest common divisor
- [X] the extended greatest common divisor

incl. tests for `Z`.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -137,3 +137,16 @@ pub trait Concatenate {
     /// if the matrices can not be concatenated due to mismatching dimensions
     fn concat_horizontal(self, other: Self) -> Result<Self::Output, MathError>;
 }
+
+/// Is implemented by [`Z`](crate::integer::Z) instances to calculate the `gcd`
+pub trait Gcd<T = Self> {
+    type Output;
+
+    /// Outputs the greatest common divisor (gcd) of the two given values.
+    ///
+    /// Paramters:
+    /// - `other`: specifies one of the values of which the gcd is computed
+    ///
+    /// Returns the greatest common divisor of `self` and `other`.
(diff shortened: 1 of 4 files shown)


Marvin-Beckmann (changes requested):
Please apply the changes from the other PR here as well

src/traits.rs
```diff
@@ -137,3 +137,33 @@ pub trait Concatenate {
...
+    /// Outputs the extended greatest common divisor (xgcd) of the two given values,
+    /// i.e. a triple `(gcd(a,b), x, y)`, where `a*x + b*y = gcd(a,b)*`.
+    ///
+    /// Paramters:
+    /// - `other`: specifies one of the values of which the gcd is computed
+    ///
+    /// Returns a triple `(gcd(a,b), x, y)` containing the greatest common divisor,
+    /// `x`, and `y` s.t. `gcd(a,b) = a*x + b*y`.
+    fn xgcd(&self, other: T) -> Self::Output;
+}
```
Marvin-Beckmann, line 169: On could consider combining this trait, since gcd and xgcd will probably not be implemented on their own, hence one can consider implementing them as a single trait with both functionalities, but I am fine with both. Feel free to do as you like here

Marvin-Beckmann (approved):
Just one general thought, which is up to your own preference

src/traits.rs
jnsiemer: Yes, I've thought about this option too. I chose to do it the separate way, because Flint also offers a `gcd` computation for `Q` (explicitely defined in the documentation), but does not offer `xgcd` computation for `Q`. This way, if we would ever want to take advantage of `Q::gcd`, the modularity of the trade would come in handy.

PR #169, opened 2023-04-09
Set_one function for MatZ

Marvin-Beckmann:
This PR adds the ability to set a matrix to the identity matrix. The function set_one sets all entries on the diagonal to `1` and the remaining entries to `0`.

- [x] set_one function for MatZ

### DIFF
--- a/src/integer/mat_z/from.rs
+++ b/src/integer/mat_z/from.rs
@@ -22,7 +22,7 @@ use crate::{
         dimensions::find_matrix_dimensions, index::evaluate_index, parse::parse_matrix_string,
     },
 };
-use flint_sys::fmpz_mat::fmpz_mat_init;
+use flint_sys::fmpz_mat::{fmpz_mat_init, fmpz_mat_one};
 use std::{fmt::Display, mem::MaybeUninit, str::FromStr};
 
 impl MatZ {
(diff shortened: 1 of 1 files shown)


src/integer/mat_z/from.rs
```diff
@@ -74,6 +74,22 @@ impl MatZ {
...
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    ///
+    /// let mut matrix = MatZ::new(2,3).unwrap();
+    /// matrix.set_one();
+    ///
+    /// let mut identity = MatZ::new(10,10).unwrap();
+    /// identity.set_one();
+    /// ```
+    pub fn set_one(&mut self) {
```
jnsiemer, line 90: As this is a `set` function, it should be in the `set` module. But it is probably a better idea to keep it in the `from` module and rename it to `identity(rows, columns)` as I see few to none use cases for setting a matrix to the identity/ one matrix after it has been set appropriately. Hence, this function would create a new matrix, set it to `one` and output it instead of manipulating it.
A bullet proof solution would be to implement both (in the appropriate modules).

jnsiemer (changes requested):

jnsiemer (approved):

PR #171, opened 2023-04-09
Toolbox for established PK encryption

jnsiemer:
This PR implements missing pieces to be able to implement established public key encryption like RSA and El-Gamal with some additional help from other libraries using our data types.

Hence, it implements...
- [x] `is_prime` for `Z`
- [x] `pow` for `Zq`

including tests.

### DIFF
--- a/src/integer/z/properties.rs
+++ b/src/integer/z/properties.rs
@@ -9,9 +9,23 @@
 //! This module includes functionality about properties of [`Z`] instances.
 
 use super::Z;
-use flint_sys::fmpz::fmpz_abs;
+use flint_sys::fmpz::{fmpz_abs, fmpz_is_prime};
 
 impl Z {
+    /// Checks if a [`Z`] is prime.
+    ///
+    /// Returns true if the value is prime.
+    ///
+    /// ```
+    /// use qfall_math::integer::Z;
+    ///
+    /// let value = Z::from(17);
+    /// assert!(value.is_prime())
+    /// ```
+    pub fn is_prime(&self) -> bool {
+        1 == unsafe { fmpz_is_prime(&self.value) }
+    }
+
     /// Computes the absolute distance between two [`Z`] instances.
     ///
     /// Parameters:
@@ -116,5 +130,28 @@ mod test_distance {
         assert_eq!(&a + Z::ONE, b.distance(&zero));
         assert_eq!(&a + Z::ONE, zero.distance(&b));
         assert_eq!(Z::ZERO, a.distance(&a));
+    }
+}
+
+#[cfg(test)]
+mod test_is_prime {
+    use super::Z;
+
+    /// ensure that primes are correctly detected
+    #[test]
+    fn prime_detection() {
+        let small = Z::from(2_i32.pow(16) + 1);
+        let large = Z::from(u64::MAX - 58);
+        assert!(small.is_prime());
+        assert!(large.is_prime());
+    }
+
+    /// ensure that non-primes are correctly detected
+    #[test]
+    fn non_prime_detection() {
+        let small = Z::from(2_i32.pow(16));
+        let large = Z::from(i64::MAX);
+        assert!(!small.is_prime());
+        assert!(!large.is_prime());
     }
 }
--- a/src/integer_mod_q/z_q/arithmetic/pow.rs
+++ b/src/integer_mod_q/z_q/arithmetic/pow.rs
@@ -1,0 +1,127 @@
+// Copyright © 2023 Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module provides an implementation of the [`Pow`] trait for [`Zq`].
+
+use crate::{
+    error::MathError,
+    integer::Z,
+    integer_mod_q::Zq,
+    macros::for_others::{implement_for_others, implement_for_owned},
+    traits::Pow,
+};
+use flint_sys::fmpz_mod::fmpz_mod_pow_fmpz;
+
+impl Pow<&Z> for &Zq {
+    type Output = Zq;
+
+    /// Raises the value of `self` to the power of an integer `exp`.
+    ///
+    /// Parameters:
+    /// - `exp`: specifies the exponent to which the value is raised
+    ///
+    /// Returns the value of `self` powered by `exp` as a new `Output` instance.
+    ///
+    /// # Example
+    /// ```
+    /// use qfall_math::integer::Z;
+    /// use qfall_math::integer_mod_q::Zq;
+    /// use qfall_math::traits::*;
+    ///
+    /// let base = Zq::try_from((2, 9)).unwrap();
+    /// let exp = Z::from(4);
+    ///
+    /// let powered_value = base.pow(&exp).unwrap();
+    ///
+    /// let cmp = Zq::try_from((7, 9)).unwrap();
+    /// assert_eq!(cmp, powered_value);
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError::InvalidExponent`] if the provided exponent is
+    /// negative and the base value of `self` is not invertible.
+    fn pow(self, exp: &Z) -> Result<Self::Output, MathError> {
+        let mut out = self.clone();
+        if 0 == unsafe {
+            fmpz_mod_pow_fmpz(
+                &mut out.value.value,
+                &self.value.value,
+                &exp.value,
+                self.modulus.get_fmpz_mod_ctx_struct(),
+            )
+        } {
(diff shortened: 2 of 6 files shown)


src/integer_mod_q/z_q/arithmetic/pow.rs
```diff
@@ -0,0 +1,127 @@
...
+            )
+        } {
+            return Err(MathError::InvalidExponent(format!(
+                "The negative exponent {} was used for a non-invertible base {}",
+                exp, self
+            )));
+        }
+        Ok(out)
+    }
+}
```
Marvin-Beckmann, line 65: Please add the powers for the other types to the kanboard and mark those you are working on

src/integer_mod_q/z_q/arithmetic/pow.rs
```diff
@@ -0,0 +1,127 @@
...
+
+implement_for_owned!(Z, Zq, Pow);
+implement_for_others!(Z, Zq, Pow for u8 u16 u32 u64 i8 i16 i32 i64);
+
+#[cfg(test)]
+mod test_pow {
+    use super::*;
+
```
Marvin-Beckmann, line 73: Please add a test for an non-invertal basis, e.g. 3 mod 9

src/integer_mod_q/z_q/arithmetic/pow.rs
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// let powered_value = base.pow(&exp).unwrap();
+    ///
+    /// let cmp = Zq::try_from((7, 9)).unwrap();
+    /// assert_eq!(cmp, powered_value);
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError::InvalidExponent`] if the provided exponent is
+    /// negative and the base value of `self` is not invertible.
```
Marvin-Beckmann, line 47:
```suggestion
    /// - Returns a [`MathError`] of type [`InvalidExponent`](MathError::InvalidExponent) if the provided exponent is
    /// negative and the base value of `self` is not invertible.
```

src/integer/z/properties.rs
```diff
@@ -118,3 +132,26 @@ mod test_distance {
         assert_eq!(Z::ZERO, a.distance(&a));
     }
 }
+
+#[cfg(test)]
+mod test_is_prime {
+    use super::Z;
+
```
Marvin-Beckmann, line 139: test for non-prime is missing

src/macros/for_others.rs
```diff
@@ -29,6 +29,8 @@
 /// - ['Mul'](std::ops::Mul) with signatures
 /// `($bridge_type:ident, $type:ident, Mul Matrix for $source_type:ident)` and
 /// `($bridge_type:ident, $type:ident, Mul Scalar for $source_type:ident)`
+/// - [`Pow`](crate::traits::Pow) with the signature
+/// `($bridge_type, $type, Pow for $source_type:ident)`
 ///
 /// # Examples
```
Marvin-Beckmann, line 35: Please also add it to the list of examples

src/macros/for_others.rs
```diff
@@ -123,6 +138,8 @@ pub(crate) use implement_for_others;
 /// `($bridge_type, $type, SetCoefficient for $source_type:ident)`
 /// - [`SetEntry`](crate::traits::SetEntry) with the signature
 /// `($bridge_type, $type, SetCoefficient for $source_type:ident)`
+/// - [`Pow`](crate::traits::Pow) with the signature
+/// `($bridge_type, $type, Pow for $source_type:ident)`
 ///
 /// # Examples
```
Marvin-Beckmann, line 144: Please also add it to the examples

src/macros/for_others.rs
```diff
@@ -109,6 +111,19 @@ macro_rules! implement_for_others {
             }
         })*
     };
+
+    // [`Pow`] trait
+    ($bridge_type:ident, $type:ident, Pow for $($source_type:ident)*) => {
+        $(impl Pow<$source_type> for &$type {
+            type Output = $type;
+            paste::paste! {
+                #[doc = "Documentation can be found at [`" $type "::pow`]. Implicitly converts [`" $source_type "`] into [`" $bridge_type "`]."]
```
Marvin-Beckmann, line 120: Link in documnetation does not work since it is not implementeed for $type but rather for &$type. While you fix it, please also fix the documentation for Mul Scalar above, since it refers to `set_entry`. In order to fix it I propose a change to the trait implementation above.

src/macros/for_others.rs
```diff
@@ -178,6 +195,22 @@ macro_rules! implement_for_owned {
             }
         }
     };
+
+    // [`Pow`] trait
+    ($source_type:ident, $type:ident, Pow) => {
+        impl Pow<$source_type> for &$type {
+            type Output = $type;
+            paste::paste! {
+                #[doc = "Documentation can be found at [`" $type "::pow`]."]
```
Marvin-Beckmann, line 204: Documentation link

src/integer_mod_q/z_q/arithmetic/pow.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn pow(self, exp: &Z) -> Result<Self::Output, MathError> {
+        let mut out = self.clone();
+        if 0 == unsafe {
```
Marvin-Beckmann, line 50: Still only takes reference, but should fix the documentation. Only left is to adapt the macro call and trait definition.
```suggestion
impl Pow<&Z> for Zq {
    type Output = Zq;

    /// Raises the value of `self` to the power of an integer `exp`.
    ///
    /// Parameters:
    /// - `exp`: specifies the exponent to which the value is raised
    ///
    /// Returns the value of `self` powered by `exp` as a new `Output` instance.
    ///
    /// # Example
    /// ```
    /// use qfall_math::integer::Z;
    /// use qfall_math::integer_mod_q::Zq;
    /// use qfall_math::traits::*;
    ///
    /// let base = Zq::try_from((2, 9)).unwrap();
    /// let exp = Z::from(4);
    ///
    /// let powered_value = base.pow(&exp).unwrap();
    ///
    /// let cmp = Zq::try_from((7, 9)).unwrap();
    /// assert_eq!(cmp, powered_value);
    /// ```
    ///
    /// # Errors and Failures
    /// - Returns a [`MathError::InvalidExponent`] if the provided exponent is
    /// negative and the base value of `self` is not invertible.
    fn pow(&self, exp: &Z) -> Result<Self::Output, MathError> {
        let mut out = self.clone();
        if 0 == unsafe {
```

Marvin-Beckmann (changes requested):
great work, just a few things that slipped through but have to be fixed

src/integer_mod_q/z_q/arithmetic/pow.rs
jnsiemer: Already done

src/integer/z/properties.rs
jnsiemer: There is a test for that in the module under the function name "non_prime_detection"

src/macros/for_others.rs
jnsiemer: It could be possible that we have multiple implementations for scalar multiplication via Phil's trait and Marcel's implementation of `mul_scalar` for MatZ. This topic should probably be discussed in the group and the macros entirely refactored after the next sprint.

src/integer_mod_q/z_q/arithmetic/pow.rs
jnsiemer: It's intentional that the base element needs to be borrowed. The syntax of `(&self).pow(exp)` is horrible to write. Hence, I rather do not give the option to pass it in as an owned type and patronize the user a bit for a much simpler way to write it `self.pow(exp)` and not get any problems afterwards. This can probably be changed in the future, if we add a special syntax like `**` or `^` to power a value, but for now I treated it for `gcd`, `xgcd`, and `distance` the same way and think it's the right design decision for now.

Marvin-Beckmann: the syntax of (&self).pow(exp) is not needed you can still write the other option, it is not enforced to do it, hence I still think my suggestion is better, since it does not change syntax allows for link highlighting in the documentation

src/integer_mod_q/z_q/arithmetic/pow.rs
Marvin-Beckmann: I was taking about Z and Q, are they added? I have not found them

jnsiemer: Yes, it's called `Pow` in `To be discussed`

src/integer_mod_q/z_q/arithmetic/pow.rs
jnsiemer: Oh, I just understood what you wanted to tell me. I'll try to figure it out.

jnsiemer: Should be done now.

Marvin-Beckmann (approved):
nice, now I'm happy :D

PR #180, opened 2023-04-12
Q `distance`

jnsiemer:
This PR implements...
- [x] distance
- [x] absolute
- [x] inv

including tests for Q.

### DIFF
--- a/src/rational/q/distance.rs
+++ b/src/rational/q/distance.rs
@@ -46,8 +46,7 @@ impl Distance<&Q> for Q {
 }
 
 implement_for_owned!(Q, Q, Distance);
-implement_for_others!(Q, Q, Distance for u8 u16 u32 u64 i8 i16 i32 i64);
-// TODO: should also be implemented for f64 once From<f64> was implemented
+implement_for_others!(Q, Q, Distance for u8 u16 u32 u64 i8 i16 i32 i64 f64 f32);
 
 #[cfg(test)]
 mod test_distance {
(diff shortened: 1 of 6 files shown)


src/rational/q/distance.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// ```
+    fn distance(&self, other: &Q) -> Self::Output {
+        let difference = other - self;
+        difference.abs()
+    }
+}
+
+implement_for_owned!(Q, Q, Distance);
+implement_for_others!(Q, Q, Distance for u8 u16 u32 u64 i8 i16 i32 i64);
+// TODO: should also be implemented for f64 once From<f64> was implemented
```
Marvin-Beckmann, line 50: this TODO can be resolved now and can also be done for f32

src/rational/q/properties.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// let value = Q::from(4);
+    ///
+    /// let inverse = value.inv().unwrap();
+    ///
+    /// assert_eq!(Q::try_from((&1, &4)).unwrap(), inverse);
+    /// ```
+    pub fn inv(&self) -> Option<Q> {
+        if self == &Q::ZERO {
+            return None;
+        }
```
Marvin-Beckmann, line 49: For this case we theoretically could use `fmpq_is_zero`. We might have to look which is more efficient, but I would  guess using the dedicated might be more efficient.

Marvin-Beckmann (changes requested):
Good work, just a quick addition

Marvin-Beckmann (approved):

PR #181, opened 2023-04-12
Zq `inv`

jnsiemer:
This PR implements `inv` for `Zq` including tests.

### DIFF
--- a/src/integer_mod_q/z_q/properties.rs
+++ b/src/integer_mod_q/z_q/properties.rs
@@ -1,0 +1,82 @@
+// Copyright © 2023 Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module includes functionality about properties of [`Zq`] instances.
+
+use super::Zq;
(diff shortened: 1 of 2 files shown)


src/integer_mod_q/z_q/properties.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// let inverse = value.inv().unwrap();
+    ///
+    /// assert_eq!(Zq::try_from((2, 7)).unwrap(), inverse);
+    /// ```
+    pub fn inv(&self) -> Option<Zq> {
+        let out = self.pow(-1);
+        if out.is_err() {
+            return None;
+        }
+        Some(out.unwrap())
```
Marvin-Beckmann, line 31:
```suggestion
        self.pow(-1).ok()
```

src/integer_mod_q/z_q/properties.rs
```diff
@@ -0,0 +1,82 @@
...
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module includes functionality about properties of [`Zq`] instances.
+
+use super::Zq;
+use crate::traits::Pow;
+
+impl Zq {
+    /// Returns the inverse of `self` as a fresh [`Zq`] instance.
+    ///
```
Marvin-Beckmann, line 16: Please add a comment that none is retruned if no inverse exists

Marvin-Beckmann (approved):
Nice work

PR #187, opened 2023-04-14
get_vec for MatZq

Marcel583:
Implements get_vec for MatZq.

### DIFF
--- a/src/integer_mod_q/mat_zq/get.rs
+++ b/src/integer_mod_q/mat_zq/get.rs
@@ -14,10 +14,11 @@ use crate::{
     integer::Z,
     integer_mod_q::{fmpz_mod_helpers::length, Modulus, Zq},
     traits::{GetEntry, GetNumColumns, GetNumRows},
-    utils::index::evaluate_indices,
+    utils::index::{evaluate_index, evaluate_indices},
 };
 use flint_sys::{
     fmpz::{fmpz, fmpz_set},
+    fmpz_mat::fmpz_mat_entry,
     fmpz_mod_mat::fmpz_mod_mat_entry,
 };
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/mat_zq/get.rs
```diff
@@ -144,7 +145,100 @@ impl GetEntry<Zq> for MatZq {
...
+    /// let col2 = matrix.get_column(2).unwrap(); // third column
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`OutOfBounds`](MathError::OutOfBounds)
+    /// if the number of the column is greater than the matrix or negative.
+    pub fn get_column(
+        &self,
+        column: impl TryInto<i64> + Display + Copy,
+    ) -> Result<Self, MathError> {
```
jnsiemer, line 220: Probably add a warning that if the matrix is reduced, the column vector will also be reduced. Otherwise, it won't, i.e. it relies on the the initial matrix to be reduced. - Same for `get_row`.

jnsiemer (changes requested):
Just one minor change regarding a warning needed

src/integer_mod_q/mat_zq/get.rs
jnsiemer: Yes, we get reduced matrices from all `from` methods, but that does not guarantee that every `MatZq` is always reduced. It's just a comment (for developers in the code) to find issues easier.

jnsiemer (approved):
If the "problem" was not filed in the refactor 6 task on the kanboard, please add it. Otherwise, this was and still is great work!

PR #188, opened 2023-04-15
Tensor product for matzq

Marvin-Beckmann:
This PR adds an implementation of the tensor product for MatZq.

- [x] tensor
- [x] tensor_safe that does returns an error rather than panic if the moduli mismatch

### DIFF
--- a/src/integer/mat_z/tensor.rs
+++ b/src/integer/mat_z/tensor.rs
@@ -26,10 +26,16 @@ impl Tensor for MatZ {
     /// use qfall_math::traits::Tensor;
     /// use std::str::FromStr;
     ///
-    /// let mat_1 = MatZ::from_str("[[1, 0, 0],[0, 1, 0],[0, 0, 1]]").unwrap();
-    /// let mat_2 = MatZ::from_str("[[1, 2, 1],[3, 4, 1]]").unwrap();
+    /// let mat_a = MatZ::from_str("[[1, 1],[2, 2]]").unwrap();
+    /// let mat_b = MatZ::from_str("[[1, 2],[3, 4]]").unwrap();
     ///
-    /// let mat_3 = mat_1.tensor_product(&mat_2);
+    /// let mat_ab = mat_a.tensor_product(&mat_b);
+    /// let res_ab = "[[1, 2, 1, 2],[3, 4, 3, 4],[2, 4, 2, 4],[6, 8, 6, 8]]";
+    /// assert_eq!(mat_ab, MatZ::from_str(res_ab).unwrap());
+    ///
+    /// let mat_ba = mat_b.tensor_product(&mat_a);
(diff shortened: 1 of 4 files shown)


src/integer_mod_q/mat_zq/tensor.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// ensure that entries are reduced by the modulus
+    #[test]
+    fn entries_reduced() {
+        let mat_1 = MatZq::from_str(&format!("[[1, 2],[3, 4]] mod {}", u64::MAX - 58)).unwrap();
+        let mat_2 = MatZq::from_str(&format!(
+            "[[1, {}],[0, {}]] mod {}",
+            u64::MAX,
+            u64::MAX - 59,
+            u64::MAX - 58
+        ))
```
MoogSven, line 258: Optional: This makes it easier to understand in my opinion.
Something similar could also be done in the other string by using `-1`, `-2` ,`-3` , `-4` instead of `{}`.
```suggestion
let mat_2 = MatZq::from_str(&format!("[[1, 58],[0, -1]] mod {}", u64::MAX - 58)).unwrap();
```

src/integer_mod_q/mat_zq/tensor.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    /// ensure that entries are reduced by the modulus
+    #[test]
+    fn entries_reduced() {
+        let mat_1 = MatZq::from_str(&format!("[[1, 2],[3, 4]] mod {}", u64::MAX - 58)).unwrap();
+        let mat_2 = MatZq::from_str(&format!(
+            "[[1, {}],[0, {}]] mod {}",
+            u64::MAX,
+            u64::MAX - 59,
+            u64::MAX - 58
```
MoogSven, line 257: Good, but I would also have been fine with a smaller modulus like 10 or so to make writing and understanding the test case easier.
Is there a reason why you use "u64::Max - **58**"? 58 feels a bit random, why not 42

MoogSven (changes requested):
Really nice work, just some minor formatting/documentation changes.

src/integer_mod_q/mat_zq/tensor.rs
Marvin-Beckmann: its a prime, no other reason :D

src/integer/mat_z/tensor.rs
```diff
@@ -32,77 +29,20 @@ impl Tensor for MatZ {
     /// let mat_1 = MatZ::from_str("[[1, 0, 0],[0, 1, 0],[0, 0, 1]]").unwrap();
     /// let mat_2 = MatZ::from_str("[[1, 2, 1],[3, 4, 1]]").unwrap();
     ///
-    /// let mat_3 = mat_1.tensor(&mat_2);
+    /// let mat_3 = mat_1.tensor_product(&mat_2);
```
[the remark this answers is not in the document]
Marvin-Beckmann: done

MoogSven (approved):
Looks good :+1:

PR #189, opened 2023-04-15
fix parse error for PolyOverZ from_str

### LINKED ISSUE #183: fix error in PolyOverZ

Marvin-Beckmann:
This closes #183.

The problem: fmpz_poly_set_str ignores both symbols after the number of coefficients, this is now fixed with a dedicated string. This issue does not occur for PolyOverQ. I added according tests in the from-methods

### DIFF
--- a/src/integer/poly_over_z/from.rs
+++ b/src/integer/poly_over_z/from.rs
@@ -61,11 +61,11 @@ impl FromStr for PolyOverZ {
             return Ok(Self::default());
         }
 
-        // fmpz_poly_set_str just skips the two symbols after the number of
-        // coefficients (even if they are not whitespaces), hence we have to check if
-        // they are whitespaces manually.
-        // and we only have to check it once, because for every other position it checks
-        // whether there is only one whitespace
+        // fmpz_poly_set_str just skips the two symbols after the first space
+        // behind the number of coefficients (even if not a space), hence
+        // it has to be checked here to ensure that no number is lost.
+        // We only have to check it once, because for every other position it checks
+        // whether there is only one space.
         if !s_trimmed.contains("  ") {
             return Err(MathError::InvalidStringToPolyMissingWhitespace(
                 s.to_owned(),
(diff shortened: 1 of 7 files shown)


src/integer/poly_over_z/from.rs
```diff
@@ -52,18 +52,30 @@ impl FromStr for PolyOverZ {
...
+
+        if s_trimmed == "0" {
+            return Ok(Self::default());
+        }
+
+        // fmpz_poly_set_str just skips the two symbols after the number of
+        // coefficients (even if they are not whitespaces), hence we have to check if
+        // they are whitespaces manually.
+        // and we only have to check it once, because for every other position it checks
+        // whether there is only one whitespace
```
MoogSven, line 66: It ensures that the first character behind the first number is a space.
Interesting Sitenote: whitespaces that are not spaces can be included within numbers, e.g. `\n`, `\t` and will be ignored.

```rust
let a = PolyOverZ::from_str("2  1\n2 3").unwrap();
let b = PolyOverZ::from_str("2  12 3").unwrap();
assert_eq!(a,b);
```

```suggestion
        // fmpz_poly_set_str just skips the two symbols after the first space
        // behind the number of coefficients (even if not a space), hence
        // it has to be checked here to ensure that no number is lost.
        // We only have to check it once, because for every other position it checks
        // whether there is only one space.
```

src/integer/poly_over_z/from.rs
```diff
@@ -52,18 +52,30 @@ impl FromStr for PolyOverZ {
     /// [`InvalidStringToCStringInput`](MathError::InvalidStringToCStringInput)
     /// if the provided string contains a Null Byte.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // remove whitespaces at the start and at the end
+        let s_trimmed = s.trim();
```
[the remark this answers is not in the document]
Marvin-Beckmann: done

src/integer/poly_over_z/from.rs
Marvin-Beckmann: interesting, I added your suggestion, do you want me to also include this sitenote?

src/rational/poly_over_q/from.rs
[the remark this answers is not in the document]
Marvin-Beckmann: done

src/integer/poly_over_z/from.rs
MoogSven: No, just the suggestion is enough. The other thing is just something interesting to know that I stumbled over during the review.

MoogSven (approved):
Looks Good :+1:

PR #190, opened 2023-04-15
Add `PartialOrd` for `Q`

MoogSven:
Test cases mostly copied from `Z`.
But changed the values slightly in the first cases to ensure that numerator and denominator are compared.

### DIFF
--- a/src/rational/q/cmp.rs
+++ b/src/rational/q/cmp.rs
@@ -68,6 +68,8 @@ impl PartialOrd for Q {
     /// assert!(a <= b);
     /// assert!(b > a);
     /// assert!(b >= a);
+    ///
+    /// assert!(&a < &b);
     /// # Ok::<(), MathError>(())
     /// ```
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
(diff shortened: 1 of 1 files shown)


src/rational/q/cmp.rs
```diff
@@ -261,3 +289,241 @@ mod test_partial_eq {
         assert_eq!(a, b);
     }
 }
+
```
3pmile, line 292: pls add a test for close to zero Q values e.g. 1/u64::MAX

src/rational/q/cmp.rs
```diff
@@ -47,6 +48,33 @@ impl PartialEq for Q {
...
+    ///
+    /// let a: Q = Q::try_from((&1,&10))?;
+    /// let b: Q = Q::try_from((&2,&10))?;
+    ///
+    /// assert!(a < b);
+    /// assert!(a <= b);
+    /// assert!(b > a);
+    /// assert!(b >= a);
+    /// # Ok::<(), MathError>(())
+    /// ```
```
3pmile, line 72: is partialOrd possible for borrowed values? if yes, plead add towards doc and tests

3pmile (changes requested):
just some minor changes

PR #194, opened 2023-04-17
gso for MatQ

Marcel583:
Implements gso for MatQ.

### DIFF
--- a/src/rational/mat_q/gso.rs
+++ b/src/rational/mat_q/gso.rs
@@ -78,7 +78,7 @@ mod test_gso {
     /// Ensure that gso works with independent vectors (more columns than rows)
     #[test]
     fn gso_dependent_columns() {
-        let mat = MatQ::from_str("[[1,2,3,4,4],[1,2,3,4,4]]").unwrap();
+        let mat = MatQ::from_str("[[1,2,3,4,4],[1,2,3,4,5]]").unwrap();
 
         let mat_gso = mat.gso();
 
--- a/src/rational/mat_q/gso.rs
+++ b/src/rational/mat_q/gso.rs
@@ -15,7 +15,7 @@ use flint_sys::fmpq_mat::fmpq_mat_gso;
 impl MatQ {
     /// Computes the Gram-Schmidt Orthogonalization of the matrix and returns a [`MatQ`] with the corresponding matrix.
     ///
-    /// # Example
+    /// # Examples
     /// ```
     /// use qfall_math::rational::MatQ;
     /// use std::str::FromStr;
@@ -77,7 +77,7 @@ mod test_gso {
 
     /// Ensure that gso works with independent vectors (more columns than rows)
     #[test]
-    fn gso_dependent_rows() {
+    fn gso_dependent_columns() {
         let mat = MatQ::from_str("[[1,2,3,4,4],[1,2,3,4,4]]").unwrap();
 
         let mat_gso = mat.gso();
@@ -96,7 +96,7 @@ mod test_gso {
 
     /// Ensure that gso works with more rows than columns
     #[test]
-    fn gso_dependent_columns() {
+    fn gso_dependent_rows() {
         let mat = MatQ::from_str("[[1,2/7],[1,2/7],[10,-2],[0,4],[0,0]]").unwrap();
 
         let mat_gso = mat.gso();
--- a/src/rational/mat_q/gso.rs
+++ b/src/rational/mat_q/gso.rs
@@ -35,34 +35,15 @@ impl MatQ {
 #[cfg(test)]
 mod test_gso {
 
-    use crate::rational::{MatQ, Q};
+    use crate::{
+        rational::{MatQ, Q},
+        traits::GetEntry,
+    };
     use std::str::FromStr;
-
-    /// Ensure that the generated vectors of the gso are orthogonal
-    #[test]
-    fn gso_orthogonal_values_over_z() {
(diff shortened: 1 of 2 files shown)


src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[test]
+    fn gso_orthogonal_values_over_q() {
+        let mat = MatQ::from_str(
+        "[[-1,2/7,3/9,4,5/2],[-123/1000,235/5,123,643/7172721,123],[124/8981,212,452/2140,12/5,1],[0,0,0,1,1],[1,2,3,4/3,1]]",
+    )
+    .unwrap();
+
+        let mat_gso = mat.gso();
+
+        let mat0 = Q::from_str("0/1").unwrap();
```
Marvin-Beckmann, line 73: once you used Q:::default() and once from_str("0/1"), please make this consistent

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[cfg(test)]
+mod test_gso {
+
+    use crate::rational::{MatQ, Q};
+    use std::str::FromStr;
+
+    /// Ensure that the generated vectors of the gso are orthogonal
+    #[test]
+    fn gso_orthogonal_values_over_z() {
```
Marvin-Beckmann, line 43: This test is not really needed imo, as its does not really differ from the toher test, as Z \subset Q

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,84 @@
...
+        unsafe {
+            fmpq_mat_gso(&mut out.matrix, &self.matrix);
+        }
+        out
+    }
+}
+
+#[cfg(test)]
+mod test_gso {
+
```
Marvin-Beckmann, line 37: please add tests for different dimensions.
What happens if some columns or rows are identical? (linear dependent)
Please add a test where some entries are large.
Please add special cases: such as a matrix with only one column, one row, etc.
What if we have more columns than rows?

Marvin-Beckmann (changes requested):

src/rational/mat_q/gso.rs
Marvin-Beckmann: * Use Q::ZERO

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,147 @@
...
+                let vec_j = mat2_gso.get_column(j).unwrap();
+                assert_eq!(cmp, vec_i.dot_product(&vec_j).unwrap());
+            }
+        }
+
+        assert_eq!(
+            MatQ::from_str("[[1],[1],[3/7]]").unwrap(),
+            mat3_gso.get_column(0).unwrap()
+        );
+    }
```
Marvin-Beckmann, line 146:
```suggestion
        assert_eq!(mat3, mat3_gso);
```

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn gso(&self) -> Self {
+        let mut out = MatQ::new(self.get_num_rows(), self.get_num_columns()).unwrap();
+        unsafe {
+            fmpq_mat_gso(&mut out.matrix, &self.matrix);
+        }
+        out
+    }
+}
+
+#[cfg(test)]
```
Marvin-Beckmann, line 35: I would like your tests to include vec_i.dot_product(&vec_j) != 0 in cases where you would expect it, and one where where you actually check how the orthogonalized matrix is what you would expect. Otherwise the function as of right now could just return a zero matrix in the correct dimension and pretty much pass all tests (except for one with [[1]])

Marvin-Beckmann (changes requested):

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,164 @@
...
+                assert_eq!(cmp, vec_i.dot_product(&vec_j).unwrap());
+            }
+        }
+        let vec1 = mat_gso.get_column(0).unwrap();
+        assert_ne!(cmp, vec1.dot_product(&vec1).unwrap());
+    }
+
+    /// Ensure that gso works with more rows than columns
+    #[test]
+    fn gso_dependent_columns() {
```
Marvin-Beckmann, line 99: these two columns are not dependent

Marvin-Beckmann (changes requested):
LOOKS GOOD, JUST TWO MINOR ADAPTIATIONS REQUIRED.

src/rational/mat_q/gso.rs
```diff
@@ -0,0 +1,164 @@
...
+        assert_eq!(
+            MatQ::from_str("[[1, -1, 1/6],[1, 0, -1/3],[1, 1, 1/6]]").unwrap(),
+            mat_gso
+        );
+    }
+
+    /// Ensure that gso works with independent vectors (more columns than rows)
+    #[test]
+    fn gso_dependent_columns() {
+        let mat = MatQ::from_str("[[1,2,3,4,4],[1,2,3,4,4]]").unwrap();
```
Marvin-Beckmann, line 81: I would change this matrix, s.t. it only has dependent columns, and not dependent rows
```suggestion
        let mat = MatQ::from_str("[[1,2,3,4,4],[1,2,3,4,5]]").unwrap();
```

Marvin-Beckmann (approved):

PR #199, opened 2023-04-20
New Feature: Rounding from Q to Z

Marvin-Beckmann:
**Description**

Rounding rationals may be necessary to reliable construct an integer from a `Q`. Rounding in general is a tool that is often used, hence adding this feature improves applicability for our library.

This PR adds the ability to round a rational `Q` to an integer value `Z`. The corresponding functionalitites are:

- [x] floor (always rounds down)
- [x] ceil (always rounds up)
- [x] round (rounds to the closest integer and if the distance is equal, then up)

**Testing**

- [x] I included assertions in which I compare expected values with computed values
- [x] I tested edge cases, where the values are positive and negative
- [x] I added tests where the value is already an integer to assure, that the value does not change when rounding

**Checklist:**

  - [x] Design choice fits code base + good time to do the PR
  - [x] The added lines provide good readability (naming, comments, documentation, spelling)
  - [x] The code fits our style guide
  - [x] My code works as intended and no side effects occur (e.g. memory leaks)
  - [x] My code is not missing any crucial feature (or if it is listed as another task)
  - [x] My code uses encapsulation and modularization

### DIFF
--- a/src/rational/q/rounding.rs
+++ b/src/rational/q/rounding.rs
@@ -1,0 +1,165 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module includes functionality about round instances of [`Q`].
+
+use super::Q;
+use crate::{integer::Z, traits::Distance};
+use flint_sys::fmpz::{fmpz_cdiv_q, fmpz_fdiv_q};
+
+impl Q {
+    /// Rounds the given rational [`Q`] down to the next integer [`Z`].
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::rational::Q;
+    /// use qfall_math::integer::Z;
+    ///
+    /// let value = Q::try_from((&5,&2)).unwrap();
+    /// assert_eq!(Z::from(2), value.floor());
+    /// ```
+    ///
+    /// ```
+    /// use qfall_math::rational::Q;
+    /// use qfall_math::integer::Z;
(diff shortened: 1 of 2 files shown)


src/rational/q/rounding.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[cfg(test)]
+mod test_floor {
+    use crate::{integer::Z, rational::Q};
+
+    // ensure that positive rationals are rounded correctly
+    #[test]
+    fn positive() {
+        let val_1 = Q::try_from((&i64::MAX, &2)).unwrap();
+        let val_2 = Q::try_from((&17, &u64::MAX)).unwrap();
```
MoogSven, line 100:
```suggestion
        let val_2 = Q::try_from((&1, &u64::MAX)).unwrap();
```
The 17 makes it more complicated then necessary. (Also applies to the next test cases)

MoogSven (changes requested):
Good Work, just small changes required.

src/rational/q/rounding.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// ```
+    /// use qfall_math::rational::Q;
+    /// use qfall_math::integer::Z;
+    ///
+    /// let value = Q::try_from((&5,&2)).unwrap();
+    /// assert_eq!(Z::from(2), value.floor());
+    ///
+    /// let value = Q::try_from((&2,&1)).unwrap();
+    /// assert_eq!(Z::from(2), value.floor());
+    /// ```
```
MoogSven, line 28:
```suggestion
    ///
    /// let value = Q::try_from((&-1,&4)).unwrap();
    /// assert_eq!(Z::from(-1), value.floor());
    /// ```
```
Please add also an example with negative values. This is to make it more clear how rounding is performed. i.e. not towards 0 (like for normal divisions) but down. This also applies to ceil (and optionally round).

MoogSven (changes requested):

PR #202, opened 2023-04-20
Foundation for uniform random sampling

jnsiemer:
This PR implements...
- [x] foundations for uniform sampling.

**Testing**

**Checklist:**

  - [x] The code provides maintainability s.t. it fulfills best practices like talking code, modularity, ...
    - [x] good readability could be discussed, but bit-wise operations are never really nice to read

### DIFF
--- a/src/utils/sample/uniform.rs
+++ b/src/utils/sample/uniform.rs
@@ -12,6 +12,7 @@
 use crate::{error::MathError, integer::Z};
 use rand::{rngs::ThreadRng, RngCore};
 
+#[allow(dead_code)]
 static mut RNG: Option<ThreadRng> = None;
 
 /// Returns a pointer to a static [`ThreadRng`] to sample uniform at random.
@@ -27,6 +28,7 @@ static mut RNG: Option<ThreadRng> = None;
 ///
 /// let rng = get_rng();
 /// ```
+#[allow(dead_code)]
 fn get_rng() -> &'static mut ThreadRng {
     if unsafe { RNG.is_none() } {
         unsafe { RNG = Some(ThreadRng::default()) };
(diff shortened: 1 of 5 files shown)


src/utils/sample/uniform.rs
```diff
@@ -0,0 +1,256 @@
...
+/// ```compile_fail
+/// use qfall_math::utils::sample::sample_uniform_rejection;
+/// use qfall_math::integer::Z;
+/// let interval_size = Z::from(20);
+///
+/// let sample = sample_uniform_rejection(&interval_size);
+///
+/// assert!(0 <= sample);
+/// assert!(sample < interval_size);
+/// ```
```
Marcel583, line 58: Does not work

src/utils/sample/uniform.rs
```diff
@@ -0,0 +1,256 @@
...
+/// assert_eq!(2, byte_vector.len());
+/// ```
+#[allow(dead_code)]
+fn sample_bits_uniform(nr_bits: usize) -> Vec<u8> {
+    let rng = get_rng();
+
+    // sample ⌈ nr_bits / 8 ⌉ bytes
+    let mut byte_vector;
+    if nr_bits % 8 == 0 {
+        byte_vector = vec![0_u8; nr_bits / 8];
```
Marcel583, line 109: why use 0_u8 here?
If it is needed, use 0u8 like in  line 118

Marcel583 (changes requested):

src/utils/sample/uniform.rs
jnsiemer: ✓

Marcel583 (approved):

PR #203, opened 2023-04-20
identity for MatZq

Marcel583:
**Description**

This PR implements identity for MatZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_zq/from.rs
+++ b/src/integer_mod_q/mat_zq/from.rs
@@ -158,7 +158,7 @@ impl MatZq {
         modulus: impl Into<Z>,
     ) -> Result<Self, MathError> {
         let mut out = MatZq::new(num_rows, num_cols, modulus)?;
-        if out.get_mod() != Modulus::try_from_z(&Z::ONE).unwrap() {
+        if Z::ONE != Z::from(&out.modulus) {
             unsafe { fmpz_mod_mat_one(&mut out.matrix) };
         }
         Ok(out)
--- a/src/integer_mod_q/mat_zq/from.rs
+++ b/src/integer_mod_q/mat_zq/from.rs
@@ -132,10 +132,11 @@ impl MatZq {
     /// Parameters:
     /// - `rum_rows`: the number of rows of the identity matrix
     /// - `num_columns`: the number of columns of the identity matrix
+    /// - `modulus`: the modulus of the matrix
     ///
     /// Returns a matrix with `1` across the diagonal and `0` anywhere else.
     ///
-    /// # Example
+    /// # Examples
     /// ```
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/mat_zq/from.rs
```diff
@@ -125,6 +125,41 @@ impl MatZq {
         }
         out
     }
+
+    /// Generate a `num_rows` times `num_columns` matrix with `1` on the
+    /// diagonal and `0` anywhere else.
```
Marvin-Beckmann, line 130: not true if modulus is `1`.

Please include that the matrix also has a modulus

src/integer_mod_q/mat_zq/from.rs
```diff
@@ -288,6 +323,71 @@ mod test_from_mat_z_modulus {
     }
 }
 
+#[cfg(test)]
+mod test_set_one {
+    use crate::{integer::Z, integer_mod_q::MatZq, traits::GetEntry};
+
```
Marvin-Beckmann, line 329: please consider the edge case with modulus `1`

Marvin-Beckmann (changes requested):

src/integer_mod_q/mat_zq/from.rs
Marcel583: That case makes the function much more inefficient

Marvin-Beckmann: interesting (I did not expect the flint methods to not check that case)... I did not expect that behaviour to be honest. I did not expect it to be this way, but since this is the case, we probably have to life with it now :D

src/integer_mod_q/mat_zq/from.rs
```diff
@@ -125,6 +125,44 @@ impl MatZq {
...
+    /// are not suited to create a matrix. For further information see [`MatZq::new`].
+    /// - Returns a [`MathError`] of type [`InvalidIntToModulus`](MathError::InvalidIntToModulus)
+    /// if the modulus is not greater than `0`. For further information see [`MatZq::new`].
+    pub fn identity(
+        num_rows: impl TryInto<i64> + Display + Copy,
+        num_cols: impl TryInto<i64> + Display + Copy,
+        modulus: impl Into<Z>,
+    ) -> Result<Self, MathError> {
+        let mut out = MatZq::new(num_rows, num_cols, modulus)?;
+        if out.get_mod() != Modulus::try_from_z(&Z::ONE).unwrap() {
```
Marvin-Beckmann, line 161: This should make it more efficient
```suggestion
        if Z::ONE != Z::from(&out.modulus) {
```

Marvin-Beckmann (approved):

PR #204, opened 2023-04-20
mul_scalar for MatQ

Marcel583:
**Description**

This PR implements mul_scalar for MatQ.

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/mat_q/arithmetic/mul_scalar.rs
+++ b/src/rational/mat_q/arithmetic/mul_scalar.rs
@@ -94,6 +94,8 @@ arithmetic_trait_borrowed_to_owned!(Mul, mul, Z, MatQ, MatQ);
 arithmetic_trait_mixed_borrowed_owned!(Mul, mul, MatQ, Z, MatQ);
 arithmetic_trait_mixed_borrowed_owned!(Mul, mul, Z, MatQ, MatQ);
 
+implement_for_others!(Q, MatQ, Mul Scalar for f32 f64);
+
 #[cfg(test)]
 mod test_mul_z {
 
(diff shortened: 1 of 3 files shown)


src/rational/mat_q/arithmetic/mul_scalar.rs
```diff
@@ -0,0 +1,301 @@
...
+        out
+    }
+}
+
+arithmetic_trait_reverse!(Mul, mul, Z, MatQ, MatQ);
+
+arithmetic_trait_borrowed_to_owned!(Mul, mul, MatQ, Z, MatQ);
+arithmetic_trait_borrowed_to_owned!(Mul, mul, Z, MatQ, MatQ);
+arithmetic_trait_mixed_borrowed_owned!(Mul, mul, MatQ, Z, MatQ);
+arithmetic_trait_mixed_borrowed_owned!(Mul, mul, Z, MatQ, MatQ);
```
jnsiemer, line 95: Please also implement `Mul<&Q> for &MatQ` for types that implement `Into<Q>` or via macro s.t. `f32` and `f64` are also possible as scalars.

src/rational/mat_q/arithmetic/mul_scalar.rs
```diff
@@ -0,0 +1,301 @@
...
+
+#[cfg(test)]
+mod test_mul_z {
```
jnsiemer, line 98: Please add a test for `availability`, i.e. a test that checks (almost) any input types that should work for this function.
You've tested a few, but if one of them shouldn't work due to changes of the implementation anymore, it should be easily detectable.

src/rational/mat_q/arithmetic/mul_scalar.rs
```diff
@@ -0,0 +1,301 @@
...
+        let integer1 = Z::from(3);
+        let integer2 = Z::from(i64::MAX);
+
+        assert_eq!(mat3, integer1 * mat1);
+        assert_eq!(mat4, integer2 * mat2);
+    }
+}
+
+#[cfg(test)]
+mod test_mul_q {
```
jnsiemer, line 208: Availability would here also be nice, if you add an implementation for `f32` and `f64`

jnsiemer (changes requested):
Good work! You should probably only extend your work a natural extension - as suggested below.

jnsiemer (approved):

PR #212, opened 2023-04-23
Z `div`

jnsiemer:
**Description**

This PR implements...
- [x] the Div trait using div_floor
- [x] div_floor, div_ceil, div_exact

for `Z`.

**Testing**

  - I decided to handle division by zero with a `panic` as it is such an obvious error and can only be caught at runtime. Hence, there would be no upside to possibly throwing an error.

**Checklist:**

### DIFF
--- a/src/integer/z/arithmetic/div.rs
+++ b/src/integer/z/arithmetic/div.rs
@@ -149,13 +149,11 @@ impl Div for &Z {
     /// let a = Z::from(42);
     /// let b = Z::from(20);
     ///
-    /// let c: Z = &a / &b;
-    /// let d: Z = a / b;
-    /// let e: Z = &c / d;
-    /// let f: Z = c / &e;
+    /// let c: Q = &a / &b;
+    /// let d: Q = a / b;
     ///
-    /// assert_eq!(Q::ONE, e);
-    /// assert_eq!(Q::try_from((2, 1)).unwrap(), f);
+    /// assert_eq!(Q::try_from((&21, &10)).unwrap(), c);
+    /// assert_eq!(Q::try_from((&21, &10)).unwrap(), d);
     /// ```
     ///
     /// # Panics ...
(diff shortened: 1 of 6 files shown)


src/integer/z/arithmetic/div.rs
```diff
@@ -0,0 +1,430 @@
...
+        assert_eq!(c, Z::from(10));
+    }
+
+    /// testing division for big [`Z`]
+    #[test]
+    fn div_large_numbers() {
+        let a: Z = Z::from(i64::MAX as u64 + 1);
+        let b: Z = Z::from(2);
+        let c: Z = Z::from(i32::MIN);
+        let d: Z = Z::from(i32::MAX as u32 + 1);
```
3pmile, line 424: not large enough, use u64 instead

src/integer/z/arithmetic/div.rs
```diff
@@ -0,0 +1,430 @@
...
+        if reminder != Z::ZERO {
+            None
+        } else {
+            Some(quotient)
+        }
+    }
+}
+
+impl Div for &Z {
+    type Output = Z;
```
3pmile, line 125: a Q would be more intuitive

src/integer/z/arithmetic/div.rs
```diff
@@ -0,0 +1,430 @@
...
+
+#[cfg(test)]
+mod test_div_floor {
+    use super::Z;
+
+    /// Checks whether `div_floor` correctly rounds non-exact quotients down
+    #[test]
+    fn floored() {
+        let small = Z::from(-11);
+        let large = Z::from(i32::MAX);
```
3pmile, line 166: i32::MAX does not excied 64 bit, is this intentional?

3pmile: please use i64::MAX/MIN, also below

3pmile (changes requested):

3pmile (approved):

PR #216, opened 2023-04-24
Benchmark RSA

jnsiemer:
**Description**

This PR implements...
- [x] a slight increase in performance of `sample_uniform`
- [x] a benchmark for RSA and the structure for further classic crypto.

As it is just a benchmark, the code does not include any tests, but is hand-tested a few times.


src/utils/sample/uniform.rs
```diff
@@ -73,14 +73,12 @@ pub(crate) fn sample_uniform_rejection(interval_size: &Z) -> Result<Z, MathError
...
-    let bit_vector = sample_bits_uniform(bit_size);
 
-    let random = Z::from(&bit_vector);
-    if &random < interval_size {
-        Ok(random)
-    } else {
-        sample_uniform_rejection(interval_size)
+    let mut random = Z::from(&sample_bits_uniform(bit_size));
+    while &random >= interval_size {
+        random = Z::from(&sample_bits_uniform(bit_size));
```
MoogSven, line 79: I like the change, but was not able to detect a statistically significant performance improvement.

benches/classic_crypto.rs
```diff
@@ -0,0 +1,104 @@
...
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Contains benchmarks for some classic cryptographic schemes.
+
+use criterion::*;
+use qfall_math::integer::Z;
+
+criterion_group!(benches, bench_rsa);
```
MoogSven, line 14: Please move this line to the bottom of the file.
Criterion also has it at the bottom of their benchmarks. (https://github.com/bheisler/criterion.rs/blob/master/benches/benchmarks/iter_with_setup.rs)

MoogSven (changes requested):
Some changes requested for readability

src/utils/sample/uniform.rs
jnsiemer: Hmm, ok. I thought so. The compiler probably optimizes the checks s.t. they do not take up any signficant time. Nevertheless, we keep the change, because it definitely does not harm the performance. Correct?

MoogSven: yes, we should keep it. I also prefer a loop over recursion, since that makes it more obvious whats going on.

benches/classic_crypto.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn static_gen() -> (Modulus, Z, Z) {
+        let modulus = Modulus::from_str(RSA_N).unwrap();
+        let pk = Z::from(65537);
+        let sk = Z::from_str(RSA_PK).unwrap();
+        (modulus, pk, sk)
+    }
+
+    /// Encrypts a `msg` assumed to be smaller than `modulus` via RSA's `pk`.
+    ///
+    /// `RSA.Enc(N, pk, msg) = msg^pk mod N`
```
MoogSven, line 89: `RSA.Enc` looks weird in rust.

```suggestion
    /// `enc(N, pk, msg) = msg^pk mod N`
```
or 
```suggestion
    /// `rsa_textbook::enc(N, pk, msg) = msg^pk mod N`
```

benches/classic_crypto.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn enc(modulus: &Modulus, pk: &Z, msg: &Z) -> Zq {
+        let msg = Zq::from_z_modulus(msg, modulus);
+        msg.pow(pk).unwrap()
+    }
+
+    /// Decrypts a `cipher` via RSA's `sk`.
+    ///
+    /// `RSA.Dec(N, sk, cipher) = cipher^sk mod N`
```
MoogSven, line 97: Same as above
```suggestion
    /// `dec(N, sk, cipher) = cipher^sk mod N`
```

benches/classic_crypto.rs
```diff
@@ -0,0 +1,133 @@
...
+/// benchmark [`rsa enc+dec`]
+pub fn bench_rsa_enc_dec(c: &mut Criterion) {
+    c.bench_function("RSA enc+dec", |b| {
+        b.iter(|| rsa_textbook::rsa_run_enc_dec())
+    });
+}
+
+/// benchmark [`rsa gen`]
+pub fn bench_rsa_gen(c: &mut Criterion) {
+    c.bench_function("RSA gen", |b| b.iter(|| rsa_textbook::rsa_run_gen()));
```
MoogSven, line 130: Why did you not just call `rsa_textbook::gen(1024)` here?
```suggestion
    c.bench_function("RSA gen", |b| b.iter(|| rsa_textbook::gen(1024)));
```

MoogSven (changes requested):

MoogSven (approved):

PR #227, opened 2023-04-28
sub Q between types

3pmile:
**Description**

This PR implements...
- [x] the Sub trait for Q - Z
- [x] the Sub trait for Q and Standard types u64, u32, u16, u8, i64, i32, i16, i8, f32, f64

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/q/arithmetic/sub.rs
+++ b/src/rational/q/arithmetic/sub.rs
@@ -9,10 +9,14 @@
 //! Implementation of the [`Sub`] trait for [`Q`] values.
 
 use super::super::Q;
-use crate::macros::arithmetics::{
-    arithmetic_trait_borrowed_to_owned, arithmetic_trait_mixed_borrowed_owned,
+use crate::{
+    integer::Z,
+    macros::arithmetics::{
+        arithmetic_between_types, arithmetic_trait_borrowed_to_owned,
+        arithmetic_trait_mixed_borrowed_owned,
+    },
 };
-use flint_sys::fmpq::fmpq_sub;
+use flint_sys::fmpq::{fmpq_sub, fmpq_sub_fmpz};
 use std::ops::Sub;
 
 impl Sub for &Q {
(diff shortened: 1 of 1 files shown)


src/rational/q/arithmetic/sub.rs
```diff
@@ -108,3 +150,178 @@ mod test_sub {
...
+        let b: u64 = 1;
+        let c: u32 = 1;
+        let d: u16 = 1;
+        let e: u8 = 1;
+        let f: i64 = 1;
+        let g: i32 = 1;
+        let h: i16 = 1;
+        let i: i8 = 1;
+        let j: f32 = 1.3;
+        let k: f64 = 1.3;
```
MoogSven, line 237: The variable names don't really explain what the value is about.
Maybe rename `a` to something like `rational` or `q_value`.
Similarly:
`b` -> `unsinged_int_64` or `uint_64`
`f` -> `int_64`
`j` -> `float_32`
etc.

MoogSven (approved):
Great work :+1: 

I would prefer it if the variable names in the last test would be more expressive than just alphabetic letters. This change should be relatively fast when using the rename feature of   VSCode (F2). However, applying it also to the other pull requests might take a bit of time.

src/rational/q/arithmetic/sub.rs
3pmile: discussed in group and current state is accepted

src/rational/q/arithmetic/sub.rs
```diff
@@ -108,3 +150,178 @@ mod test_sub {
...
+    use std::str::FromStr;
+
+    /// testing subtraction for [`Q`] and [`Z`]
+    #[test]
+    fn sub() {
+        let a: Q = Q::from_str("5/7").unwrap();
+        let b: Z = Z::from(4);
+        let c: Q = a - b;
+        assert_eq!(c, Q::from_str("-23/7").unwrap());
+    }
```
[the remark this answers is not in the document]
3pmile: discussed in group and current state is accepted

PR #228, opened 2023-04-29
trace for MatZ

Marcel583:
**Description**

This PR implements trace for MatZ.

**Testing**

**Checklist:**

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -56,7 +56,7 @@ use thiserror::Error;
 /// - `NotNaturalNumber` is thrown if the function expects a natural number,
 /// but a number smaller than `1` is provided
 /// - `NotPrime` is thrown if a provided integer is not prime
-/// - `NotSquare` is thrown if a matrix is not square
+/// - `NoSquareMatrix` is thrown if a matrix is not square
 /// - `OutOfBounds` is thrown if a provided index is not in a desired range
 /// - `VectorFunctionCalledOnNonVector` is thrown if a function defined
 /// on vectors was called on a matrix instance that is not a vector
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -56,6 +56,7 @@ use thiserror::Error;
 /// - `NotNaturalNumber` is thrown if the function expects a natural number,
 /// but a number smaller than `1` is provided
 /// - `NotPrime` is thrown if a provided integer is not prime
+/// - `NotSquare` is thrown if a matrix is not square
```
3pmile, line 59:
```suggestion
/// - `NoSquareMatrix` is thrown if a matrix is not square
```

would be easier to understand

src/error.rs
```diff
@@ -180,6 +181,10 @@ pub enum MathError {
     #[error("invalid integer. The integer has to be prime and the provided value is {0}")]
     NotPrime(String),
 
+    /// if a matrix is not square
+    #[error("the matrix is not square {0}")]
+    NotSquare(String),
```
3pmile, line 186:
```suggestion
    NoSquareMatrix(String),
```

see above

src/integer/mat_z/trace.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl MatZ {
+    /// Returns the trace of a matrix and an error,
+    /// if the matrix is not square.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    /// use qfall_math::traits::*;
+    /// use std::str::FromStr;
```
3pmile, line 28:
```suggestion
    /// use qfall_math::integer::MatZ;
    /// use std::str::FromStr;
```

this import is not used

3pmile (changes requested):

3pmile (approved):

PR #252, opened 2023-05-07
Refactor sprint 6 (Partial)

MoogSven:
**Description**
Each fix is a single commit with an appropriate commit message. I strongly suggest to review the changes commit by commit and not all at once.

Please have a closer look at the changes to evaluate_index  [go to changes](https://github.com/qfall/math/commit/3436e837d3c43145423ed732e2655b232c54e887#diff-5ea21bbe82d1ea8d43ea2c0121b0fd73e503c31e20b1524d63a43e2d97d48fee)

This PR implements parts of the refactoring of sprint 6
- [x] move `z/logarithm.rs` to `z/arithmetic/logarithm.rs`, exponentiation, root
- [x] `impl TryInto<i64> + Display + Copy` does not allow `Z` since `Z` does not implement Copy => Try to do it without Copy, maybe replace Copy with Clone
- [x] change '0' to `0` in docs (4 cases)
- [x] Identity and new in Matrices to Default (default.rs)
- [x] rename module information in get.rs of all matrices to (... get information about ...)

**Testing**
Just stuff that is also tested in the pipeline.

**Checklist:**


Marvin-Beckmann (commented):
Nice work, i still get some warnings, can you fix them in this PR?:D
Can you fix these warnings?
image

src/utils/index.rs
```diff
@@ -31,15 +31,29 @@ use std::fmt::Display;
...
     // the index must fit into an [`i64`]
-    let index: i64 = match index.try_into() {
-        Ok(index) => index,
-        _ => {
-            return Err(MathError::OutOfBounds(
-                "fit into a i64".to_owned(),
-                index.to_string(),
-            ))
+
+    let index: i64 = if cfg!(debug_assertions) {
```
Marvin-Beckmann, line 37: please also add a comment in our code that we differentiate between debug execution and normal excecution

src/utils/index.rs
```diff
@@ -31,15 +31,29 @@ use std::fmt::Display;
...
+        let index_str = index.to_string();
+        match index.try_into() {
+            Ok(index) => index,
+            _ => {
+                return Err(MathError::OutOfBounds(
+                    "fit into a i64".to_owned(),
+                    index_str,
+                ))
+            }
```
Marvin-Beckmann, line 46: it feel like this can be shorter, but I failed to find a better way. It would be nice if the tryInto returns a matherrror, if we would be able to propagate it

Marvin-Beckmann (changes requested):
nice work

src/utils/index.rs
```diff
@@ -31,15 +31,29 @@ use std::fmt::Display;
 /// # Errors and Failures
 /// - Returns a [`MathError`] of type [`OutOfBounds`](MathError::OutOfBounds) if
 /// either the index is negative or it does not fit into an [`i64`].
-pub fn evaluate_index<S: TryInto<i64> + Display + Copy>(index: S) -> Result<i64, MathError> {
+pub fn evaluate_index<S: TryInto<i64> + Display>(index: S) -> Result<i64, MathError> {
```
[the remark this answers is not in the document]
MoogSven: added

src/utils/index.rs
MoogSven: added

src/utils/index.rs
MoogSven: Also didn't find a good way to do this :disappointed:

MoogSven:
> Nice work, i still get some warnings, can you fix them in this PR?:D Can you fix these warnings? image

No, that can be done in the next refactoring.

Marvin-Beckmann (approved):

PR #258, opened 2023-05-11
MatZ `swap` + `reverse`

jnsiemer:
**Description**

This PR implements...
- [x] swap columns and rows
- [x] reverse columns and rows

for `MatZ`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_poly_over_z/set.rs
+++ b/src/integer/mat_poly_over_z/set.rs
@@ -24,6 +24,9 @@ impl SetEntry<&PolyOverZ> for MatPolyOverZ {
     /// - `column`: specifies the column in which the entry is located
     /// - `value`: specifies the value to which the entry is set
     ///
+    /// Returns an empty `Ok` if the action could be performed successfully.
+    /// Otherwise, a [`MathError`] is returned if the specified entry is not part of the matrix.
+    ///
     /// # Examples
     /// ```
     /// use qfall_math::integer::MatPolyOverZ;
--- a/src/integer/mat_z/set.rs
+++ b/src/integer/mat_z/set.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Marcel Luca Schmidt
+// Copyright © 2023 Marcel Luca Schmidt, Niklas Siemer
 //
 // This file is part of qFALL-math.
 //
@@ -35,6 +35,9 @@ impl SetEntry<&Z> for MatZ {
     /// - `row`: specifies the row in which the entry is located
     /// - `column`: specifies the column in which the entry is located
     /// - `value`: specifies the value to which the entry is set
+    ///
+    /// Returns an empty `Ok` if the action could be performed successfully.
+    /// Otherwise, a [`MathError`] is returned if the specified entry is not part of the matrix.
     ///
     /// # Examples
     /// ```
--- a/src/integer/mat_z/set.rs
+++ b/src/integer/mat_z/set.rs
@@ -13,11 +13,11 @@ use crate::{
     error::MathError,
     integer::Z,
     macros::for_others::{implement_for_others, implement_for_owned},
-    traits::{GetEntry, GetNumColumns, GetNumRows, SetEntry},
+    traits::{GetNumColumns, GetNumRows, SetEntry},
     utils::index::{evaluate_index, evaluate_indices},
 };
 use flint_sys::{
-    fmpz::fmpz_set,
+    fmpz::{fmpz_set, fmpz_swap},
     fmpz_mat::{
         fmpz_mat_entry, fmpz_mat_invert_cols, fmpz_mat_invert_rows, fmpz_mat_swap_cols,
         fmpz_mat_swap_rows,
@@ -111,11 +111,12 @@ impl MatZ {
         let (row0, col0) = evaluate_indices(self, row0, col0)?;
         let (row1, col1) = evaluate_indices(self, row1, col1)?;
 
-        let entry0 = self.get_entry(row0, col0)?;
-        let entry1 = self.get_entry(row1, col1)?;
-
-        self.set_entry(row0, col0, entry1)?;
-        self.set_entry(row1, col1, entry0)?;
+        unsafe {
+            fmpz_swap(
+                fmpz_mat_entry(&self.matrix, row0, col0),
+                fmpz_mat_entry(&self.matrix, row1, col1),
+            )
+        };
         Ok(())
     }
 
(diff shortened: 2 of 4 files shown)


src/integer/mat_z/set.rs
```diff
@@ -62,6 +78,118 @@ implement_for_owned!(Z, MatZ, SetEntry);
...
+                },
+            ));
+        }
+        unsafe { fmpz_mat_swap_rows(&mut self.matrix, null(), row0, row1) }
+        Ok(())
+    }
+}
+
+impl MatZ {
+    /// Reverses all columns of the specified matrix.
```
Marvin-Beckmann, line 166: not quite clear what this means, maybe add one more sentence for clearification

src/integer/mat_z/set.rs
```diff
@@ -62,6 +78,118 @@ implement_for_owned!(Z, MatZ, SetEntry);
...
+    /// use qfall_math::integer::MatZ;
+    ///
+    /// let mut matrix = MatZ::new(4, 3).unwrap();
+    /// matrix.reverse_columns();
+    /// ```
+    pub fn reverse_columns(&mut self) {
+        unsafe { fmpz_mat_invert_cols(&mut self.matrix, null_mut()) }
+    }
+
+    /// Reverses all rows of the specified matrix.
```
Marvin-Beckmann, line 179: See above

src/integer/mat_z/set.rs
```diff
@@ -62,6 +78,118 @@ implement_for_owned!(Z, MatZ, SetEntry);
...
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    ///
+    /// let mut matrix = MatZ::new(4, 3).unwrap();
+    /// matrix.reverse_columns();
+    /// ```
+    pub fn reverse_columns(&mut self) {
+        unsafe { fmpz_mat_invert_cols(&mut self.matrix, null_mut()) }
```
Marvin-Beckmann, line 176: why can we provide null_mut() as a second parameter. Please add a comment, such that developers can understand

src/integer/mat_z/set.rs
```diff
@@ -177,3 +305,181 @@ mod test_setter {
...
+    #[test]
+    fn row_out_of_bounds() {
+        let mut matrix = MatZ::new(2, 4).unwrap();
+
+        assert!(matrix.swap_rows(-1, 0).is_err());
+        assert!(matrix.swap_rows(0, -1).is_err());
+        assert!(matrix.swap_rows(4, 0).is_err());
+        assert!(matrix.swap_rows(0, 4).is_err());
+    }
+}
```
Marvin-Beckmann, line 407: can you add a swap for an equal column and row, i.e. swap_rows/cols(1,1)

src/traits.rs
```diff
@@ -204,3 +204,43 @@ pub trait Xgcd<T = Self> {
...
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`OutOfBounds`](MathError::OutOfBounds)
+    /// if one of the given rows is greater than the matrix or negative.
+    fn swap_rows(
+        &mut self,
+        row0: impl TryInto<i64> + Display,
+        row1: impl TryInto<i64> + Display,
+    ) -> Result<(), MathError>;
+}
```
Marvin-Beckmann, line 246: I would expect the functionality swap_entries alongside the functions presented by you.
Consider adding it, or adding a corresponding task to the kanboard for discussion

src/integer/mat_z/set.rs
```diff
@@ -62,6 +78,118 @@ implement_for_owned!(Z, MatZ, SetEntry);
 
 implement_for_others!(Z, MatZ, SetEntry for i8 i16 i32 i64 u8 u16 u32 u64);
 
+impl MatrixSwaps for MatZ {
```
Marvin-Beckmann, line 81: swap to me implies swaps with extern matrices, I would prefer naming like `arrange, rearrange`, but swap is probably also fine. But iI would use the singular for the trait name to stay consistent

Marvin-Beckmann (changes requested):
Nice additions, please consider my suggestions

src/integer/mat_z/set.rs
jnsiemer: I removed the trait now that it is not needed anymore for sorting. If needed anytime, we can always re-introduce it.
I sticked to the word "swap" as I believe that it is the more intuitive word to search for than "rearrange" or "arrange". You're correct that this should probably include also swapping elements with other matrices, but it would add at least one parameter to every function, which does not make the API any nicer.
Hence, I would stick to `get_column` and `set_column` (which will be implemented now) for inter-matrix swaps for now.

src/traits.rs
jnsiemer: Functionality was added, but is to my eye not much nicer than just manually using `get_entry` and `set_entry` in one line

src/integer/mat_poly_over_z/set.rs
```diff
@@ -24,6 +24,9 @@ impl SetEntry<&PolyOverZ> for MatPolyOverZ {
     /// - `column`: specifies the column in which the entry is located
     /// - `value`: specifies the value to which the entry is set
     ///
+    /// Returns an empty `Ok` if the action could be performed successfully.
+    /// Otherwise, a [`MathError`] is returned if the specified entry is not part of the matrix.
+    ///
```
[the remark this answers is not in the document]
jnsiemer: Added to refactor 7

src/integer/mat_z/set.rs
```diff
@@ -62,6 +78,164 @@ implement_for_owned!(Z, MatZ, SetEntry);
...
+        col1: impl TryInto<i64> + Display,
+    ) -> Result<(), MathError> {
+        let (row0, col0) = evaluate_indices(self, row0, col0)?;
+        let (row1, col1) = evaluate_indices(self, row1, col1)?;
+
+        let entry0 = self.get_entry(row0, col0)?;
+        let entry1 = self.get_entry(row1, col1)?;
+
+        self.set_entry(row0, col0, entry1)?;
+        self.set_entry(row1, col1, entry0)?;
```
Marvin-Beckmann, line 118: should there be a more efficient possible implementation, where we can use the FLINT methods for this, using fmpz_mat_entry and then just swapping the values (i.e. the pointer or respectively the value). It is not as important for Z, but looking forward to other structs like MatPolyZ or similiar this could be a great imporvement.

If you do not want to improve it now, just add a todo or similar.

Marvin-Beckmann (approved):

PR #260, opened 2023-05-11
norms poly over z

3pmile:
**Description**

This PR implements...
- [x] squared  euclidean norm
- [x] infinity norm
- [x] get degree

for/ of `PolyOverZ`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/poly_over_z/get.rs
+++ b/src/integer/poly_over_z/get.rs
@@ -11,7 +11,7 @@
 
 use super::PolyOverZ;
 use crate::{error::MathError, integer::Z, traits::GetCoefficient, utils::index::evaluate_index};
-use flint_sys::fmpz_poly::fmpz_poly_get_coeff_fmpz;
+use flint_sys::fmpz_poly::{fmpz_poly_degree, fmpz_poly_get_coeff_fmpz};
 use std::fmt::Display;
 
 impl GetCoefficient<Z> for PolyOverZ {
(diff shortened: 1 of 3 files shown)


src/integer/poly_over_z/get.rs
```diff
@@ -49,6 +49,24 @@ impl GetCoefficient<Z> for PolyOverZ {
...
+    ///
+    /// let poly = PolyOverZ::from_str("4  0 1 2 3").unwrap();
+    ///
+    /// let degree = poly.get_degree(); // This would only return 3
+    /// ```
+    pub fn get_degree(&self) -> i64 {
+        unsafe { fmpz_poly_degree(&self.poly) }
+    }
+}
+
```
Marvin-Beckmann, line 69: I would open the discussion whether this should be in get.rs or in properties.rs. We have introduced properties.rs for some other types, I personally think, that it could fit into both, what do you think?
Probably get is fine, but lets just briefly talk about it to make it consistently

Marvin-Beckmann: alternatively you could have considered adding length, which returns the number of coefficients, was this a purposeful decision to choose degree over length? Because I think that calling the length function within the norms would make the code readable a little better

src/integer/poly_over_z/norm.rs
```diff
@@ -0,0 +1,139 @@
...
+    ///
+    /// let sqrd_2_norm = poly.norm_eucl_sqrd();
+    ///
+    /// // 1*1 + 2*2 + 3*3 = 14
+    /// assert_eq!(Z::from(14), sqrd_2_norm);
+    /// ```
+    pub fn norm_eucl_sqrd(&self) -> Z {
+        let mut res = Z::ZERO;
+        for i in 0..=self.get_degree() {
+            res = res + self.get_coeff(i).unwrap() * self.get_coeff(i).unwrap();
```
Marvin-Beckmann, line 35:
```suggestion
            res = res + self.get_coeff(i).unwrap().pow(2).unwrap();
```

Marvin-Beckmann (changes requested):
looks good, just one topic regarding style and readability

src/integer/poly_over_z/norm.rs
Marvin-Beckmann: pow is implemented

src/integer/poly_over_z/get.rs
Marvin-Beckmann: what do you think about the second question?

3pmile: for now i just wanted to implement on of the two functions and get_degree is more intuitive to use by the end user

src/integer/poly_over_z/norm.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        for i in 0..=self.get_degree() {
+            let coeff = self.get_coeff(i).unwrap();
+            res = res + coeff.pow(2).unwrap();
+        }
+        res
+    }
+}
+
+impl PolyOverZ {
```
Marvin-Beckmann, line 43:
```suggestion

```

Marvin-Beckmann (approved):

PR #264, opened 2023-05-12
Ensure that dev caches can be found by pull requests

jnsiemer:


.github/workflows/main.yml
```diff
@@ -84,7 +84,6 @@ jobs:
       
       - name: Archive release build
         uses: actions/upload-artifact@v1
-        if: github.ref == 'ref/head/master'
         with:
           name: ${{ env.PROJECT_NAME }}-release_build-v${{ env.PROJECT_VERSION }}
           path: target/release/${{ env.PROJECT_NAME }}
```
Marvin-Beckmann, line 89: wouldn't this mean that after every push on dev a release version is build?

.github/workflows/pull_request.yml
```diff
@@ -1,7 +1,13 @@
...
 # consistency regarding formatting and idiomatic Rust
 
-on: [ pull_request ]
+on: 
+  push:
+    branches:
+      - dev
+  pull_request:
+    branches:
+      - "**"
```
Marvin-Beckmann, line 10: would this mean that the cash is used for every single branch? Even if we lets say realize that we have to reset the branch to an older state, would this mean that the cache would still be used for this PR?

Marvin-Beckmann (changes requested):
I am not really shure why this fixes our issue, can you provide an intuition why it does. Please add it to the description of the PR

.github/workflows/main.yml
MoogSven: no, because it is now just running on main (first change)

MoogSven (approved):
looks good :+1:

PR #266, opened 2023-05-13
norm poly over Zq

3pmile:
**Description**

This PR implements...
- [x] squared  euclidean norm
- [x] infinity norm
- [x] get degree

for/ of `PolyOverZq`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/poly_over_zq/get.rs
+++ b/src/integer_mod_q/poly_over_zq/get.rs
@@ -108,7 +108,7 @@ impl PolyOverZq {
     /// let degree = poly.get_degree(); // This would only return 3
     /// ```
     pub fn get_degree(&self) -> i64 {
-        unsafe { fmpz_mod_poly_degree(&self.poly, &*self.modulus.modulus) }
+        unsafe { fmpz_mod_poly_degree(&self.poly, self.modulus.get_fmpz_mod_ctx_struct()) }
     }
 }
 
--- a/src/integer_mod_q/poly_over_zq/get.rs
+++ b/src/integer_mod_q/poly_over_zq/get.rs
@@ -14,7 +14,7 @@ use crate::{
     error::MathError, integer::Z, integer_mod_q::Zq, traits::GetCoefficient,
     utils::index::evaluate_index,
 };
-use flint_sys::fmpz_mod_poly::fmpz_mod_poly_get_coeff_fmpz;
+use flint_sys::fmpz_mod_poly::{fmpz_mod_poly_degree, fmpz_mod_poly_get_coeff_fmpz};
 use std::fmt::Display;
 
 impl GetCoefficient<Zq> for PolyOverZq {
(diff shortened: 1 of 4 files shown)


src/integer_mod_q/poly_over_zq/get.rs
```diff
@@ -94,6 +94,24 @@ impl GetCoefficient<Z> for PolyOverZq {
...
+    /// ```
+    /// use qfall_math::integer_mod_q::PolyOverZq;
+    /// use std::str::FromStr;
+    ///
+    /// let poly = PolyOverZq::from_str("4  0 1 2 3 mod 7").unwrap();
+    ///
+    /// let degree = poly.get_degree(); // This would only return 3
+    /// ```
+    pub fn get_degree(&self) -> i64 {
+        unsafe { fmpz_mod_poly_degree(&self.poly, &*self.modulus.modulus) }
```
3pmile, line 111: @MoogSven does this lead to a memory leak?

src/integer_mod_q/poly_over_zq/get.rs
```diff
@@ -183,3 +201,54 @@ mod test_get_coeff_z {
...
+        let deg = poly.get_degree();
+
+        assert_eq!(0, deg);
+    }
+
+    /// ensure that degree is working for polynomials with many coefficients
+    /// flint does not reduce the exponent due to computational cost
+    #[test]
+    fn degree_many_coefficients() {
+        let poly = PolyOverZq::from_str("7  1 2 3 4 8 1 3 mod 2").unwrap();
```
Marvin-Beckmann, line 248: can you make this modulus large, e.g. u64::MAX?
and one with a large coefficient

src/integer_mod_q/poly_over_zq/norm.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn norm_eucl_sqrd(&self) -> Z {
+        let mut res = Z::ZERO;
+        for i in 0..=self.get_degree() {
+            let mut coeff: Z = self.get_coeff(i).unwrap();
+            let minus_coeff = Z::from(&self.modulus) - &coeff;
+            if minus_coeff < coeff {
+                //todo: use min once on dev
+                coeff = minus_coeff;
+            }
+            res = res + coeff.pow(2).unwrap();
```
Marvin-Beckmann, line 42: rather use the distance function for Zq values, I think this would improve readability here

src/integer_mod_q/poly_over_zq/norm.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// ```
+    pub fn norm_infty(&self) -> Z {
+        let mut res = Z::ZERO;
+        for i in 0..=self.get_degree() {
+            let mut coeff: Z = self.get_coeff(i).unwrap();
+            let minus_coeff = Z::from(&self.modulus) - &coeff;
+            if minus_coeff < coeff {
+                //todo: use min once on dev
+                coeff = minus_coeff;
+            }
```
Marvin-Beckmann, line 72: see above

Marvin-Beckmann (changes requested):

Marvin-Beckmann (changes requested):

src/integer_mod_q/poly_over_zq/get.rs
MoogSven: There is no memory leak. However, it could be simplified to:
```rs
self.poly.length - 1
```
This is also what happens in FLINT.

Marvin-Beckmann (approved):

PR #268, opened 2023-05-15
Rework qfromf64

Marvin-Beckmann:
**Description**

The current implementation of `Q::from_f64` was falty. This PR fixes it, such that all floats can be entered.
The new implementation looks at the value in memory, and takes the exact valeu represented. This means that values which can not be represented perfectly, will only use the bit represenatation from memory.
Subsequent changes had to be made to older tests, which either rounded or used lower precision.
This PR also updates the constant values for `e` and `pi` to 62 bit representation in denominator and numerator (more not possible due to Flints integer representation.

This PR implements...
- [x] bugfix for `Q::from_f64`

**Testing**

- [x] I adapted old tests
- [x] I added new tests for long floats

**Checklist:**

### DIFF
--- a/src/rational/q/from.rs
+++ b/src/rational/q/from.rs
@@ -199,9 +199,9 @@ impl Q {
     }
 
     /// Create a new rational number of type [`Q`] from a [`f64`].
-    /// This function works with the bit representation of the float it received as input.
-    /// Floats like `0.1` that are not completely representable,
-    /// will not be instantiated as `1/10`.
+    /// This function works with the exact float it received as input.
+    /// Many numbers like `0.1` are not exactly representable as floats and 
+    /// will therefore not be instantiated as `1/10`.
     ///
     /// Input parameters:
     /// - `value` : The value the rational number will have, provided as a [`f64`]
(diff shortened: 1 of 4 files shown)


src/rational/q/from.rs
```diff
@@ -213,15 +216,29 @@ impl Q {
...
+            // prepend one bit to the mantissa because the most significant bit is implicit
+            (bits & 0xfffffffffffff) | 0x10000000000000
+        };
+
+        // -1023 because of the offset representation of the exponent
+        // -52 because the mantissa is 52 bit long
+        exponent -= 1023 + 52;
+        let shift = match exponent {
+            e if e >= 1 => Q::from(2).pow(e).unwrap(),
+            e => Q::try_from((&1, &2)).unwrap().pow(e.abs()).unwrap(),
```
MoogSven, line 238:
```suggestion
            // This could be optimized with `fmpz_lshift_mpn` once it is part of flint_sys.
            e if e >= 1 => Q::from(2).pow(e).unwrap(),
            e => Q::try_from((&1, &2)).unwrap().pow(e.abs()).unwrap(),
```
This is not super efficient, but just add a comment for now.

MoogSven (approved):
Looks good, just small changes :+1:

PR #276, opened 2023-05-17
Add for MatPolyRingZq

Marcel583:
**Description**

This PR implements Add for MatPolyRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
+++ b/src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
@@ -11,11 +11,9 @@
 use super::super::MatPolynomialRingZq;
 use crate::{
     error::MathError,
-    integer::MatPolyOverZ,
     macros::arithmetics::{
         arithmetic_trait_borrowed_to_owned, arithmetic_trait_mixed_borrowed_owned,
     },
-    traits::{GetNumColumns, GetNumRows},
 };
 use std::ops::Add;
 
@@ -50,6 +48,7 @@ impl Add for &MatPolynomialRingZq {
     ///
     /// # Panics ...
     /// - ... if the moduli of both [`MatPolynomialRingZq`] mismatch.
+    /// - ... if the dimensions of both [`MatPolynomialRingZq`] mismatch.
     fn add(self, other: Self) -> Self::Output {
         self.add_safe(other).unwrap()
     }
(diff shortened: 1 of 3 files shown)


src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// let poly_ring_mat2 = MatPolynomialRingZq::from((&poly_mat2, &modulus));
+    ///
+    /// let poly_ring_mat3: MatPolynomialRingZq = &poly_ring_mat1 + &poly_ring_mat2;
+    /// let poly_ring_mat4: MatPolynomialRingZq = poly_ring_mat1 + poly_ring_mat2;
+    /// let poly_ring_mat5: MatPolynomialRingZq = &poly_ring_mat3 + poly_ring_mat4;
+    /// let poly_ring_mat6: MatPolynomialRingZq = poly_ring_mat3 + &poly_ring_mat5;
+    /// ```
+    ///
+    /// # Panics ...
+    /// - ... if the moduli of both [`MatPolynomialRingZq`] mismatch.
```
Marvin-Beckmann, line 52: panic if dimensions mismatch, is missing

src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
```diff
@@ -0,0 +1,278 @@
...
+#[cfg(test)]
+mod test_add {
+
+    use crate::integer::MatPolyOverZ;
+    use crate::integer_mod_q::MatPolynomialRingZq;
+    use crate::integer_mod_q::ModulusPolynomialRingZq;
+    use std::str::FromStr;
+
+    const BITPRIME64: u64 = 18446744073709551557;
+
```
Marvin-Beckmann, line 127: tests for different dimensions is missing

src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// both [`MatPolynomialRingZq`] mismatch.
+    pub fn add_safe(&self, other: &Self) -> Result<MatPolynomialRingZq, MathError> {
+        if self.modulus != other.modulus {
+            return Err(MathError::MismatchingModulus(format!(
+                " Tried to add polynomial with modulus '{}' and polynomial with modulus '{}'.",
+                self.modulus, other.modulus
+            )));
+        }
+        let mut out = MatPolynomialRingZq::from((
+            &MatPolyOverZ::new(self.get_num_rows(), self.get_num_columns()).unwrap(),
```
Marvin-Beckmann, line 93: it would be nice, if there is an implementation, that takes this in as owned, because it would allow for a more efficient implementation. Please add a todo AND an issue/or task, that an owned from is implemented, because an entire copy would otherwise be done without need.

Should also be added for the PolyOverZ, but here it is an even greater loss of effciency. Fix it later, but add an issue, that we do not forget

Marvin-Beckmann (changes requested):

src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
Marcel583: This is already implemented (see macros)

Marvin-Beckmann (approved):

PR #278, opened 2023-05-17
Generic `sort` + MatZ `sort`

jnsiemer:
**Description**

This PR implements...
- [x] sort_by_column for a generic function performed on each column vector
- [x] sort_by_row for a generic function performed on each row vector

for `MatZ`.

- [x] generic quicksort algorithm using a `cond_vec` to sort the values by and an `id_vec` to provide the permutation after sorting acc. to that certain condition

**Testing**

**Checklist:**


src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+use super::MatZ;
+use crate::{
+    error::MathError,
+    traits::{GetNumColumns, GetNumRows},
+    utils::sort::quicksort,
+};
+
+impl MatZ {
+    #[allow(clippy::needless_range_loop)]
+    /// Sorts the matrix based on some via `cond_func` specified condition of its columns.
```
Marvin-Beckmann, line 20:
```suggestion
    /// Sorts the columns of matrix based on some condition defined by `cond_func`.
```

src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        Ok(out)
+    }
+
+    #[allow(clippy::needless_range_loop)]
+    /// Sorts the matrix based on some via `cond_func` specified condition of its rows.
+    ///
+    /// This condition is usually a norm with the described input-output behaviour.
+    ///
+    /// Paramters:
```
Marvin-Beckmann, line 94:
```suggestion
    /// Parameters:
```

src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+    utils::sort::quicksort,
+};
+
+impl MatZ {
+    #[allow(clippy::needless_range_loop)]
+    /// Sorts the matrix based on some via `cond_func` specified condition of its columns.
+    ///
+    /// This condition is usually a norm with the described input-output behaviour.
+    ///
+    /// Paramters:
```
Marvin-Beckmann, line 24:
```suggestion
    /// Parameters:
```

src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+            lengths.push(cond_func(&self.get_column(col).unwrap())?);
+        }
+        let length = lengths.len() - 1;
+        let mut id_vec: Vec<usize> = (0..self.get_num_columns() as usize).collect();
+        quicksort(&mut lengths, 0, length, &mut id_vec);
+
+        let mut out = Self::new(self.get_num_rows(), self.get_num_columns()).unwrap();
+        for col in 0..id_vec.len() {
+            out.set_column(col, self, id_vec[col]).unwrap();
+        }
```
Marvin-Beckmann, line 84: Rather do something like this than surpress the clippy warning. Also for the other case
```suggestion
        for (col, elem) in id_vec.iter().enumerate() {
            out.set_column(col, self, *elem).unwrap();
        }
```

src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+    where
+        T: PartialOrd,
+    {
+        let mut lengths = vec![];
+        for col in 0..self.get_num_columns() {
+            lengths.push(cond_func(&self.get_column(col).unwrap())?);
+        }
+        let length = lengths.len() - 1;
+        let mut id_vec: Vec<usize> = (0..self.get_num_columns() as usize).collect();
+        quicksort(&mut lengths, 0, length, &mut id_vec);
```
Marvin-Beckmann, line 79: This removes the need to use the self-implemented quicksort and uses rusts timsort, which is build in. This only requires use to implement Ord for Z, which we wanted to do anyway, and *should* not be too much.
As this would reduce our code basis by ~ 200 lines. I tested it locally, just change PartialOrd to Ord as a condition and implement Ord for Z and everything works out perfectly fine, i.e. all tests work
```suggestion
        let mut id_vec: Vec<usize> = (0..self.get_num_columns() as usize).collect();
        id_vec.sort_by_key(|x| &lengths[*x]);
```

Marvin-Beckmann: additionally update the corresponding documentation and the sort by row function as well

src/integer/mat_z/sort.rs
```diff
@@ -0,0 +1,296 @@
...
+        let mut out = Self::new(self.get_num_rows(), self.get_num_columns()).unwrap();
+        for col in 0..id_vec.len() {
+            out.set_column(col, self, id_vec[col]).unwrap();
+        }
+
+        Ok(out)
+    }
+
+    #[allow(clippy::needless_range_loop)]
+    /// Sorts the matrix based on some via `cond_func` specified condition of its rows.
```
Marvin-Beckmann, line 90: see above

Marvin-Beckmann (changes requested):
I do not think, we should implement quicksort ourselves. I proposed an implementation with the rust buildins. This only requires us ti implement Ord for Z, which we wanted to anyway if I am not mistaken. Hence I would like to see these changes included.
Generally nice work, I just think, that we can make our life much easier here :D

Marvin-Beckmann (approved):

PR #285, opened 2023-05-24
fixed issue

Marcel583:
**Description**

This PR fixes the issue that only primes could be entered into a modulus for polynomial rings.

**Testing**

**Checklist:**


Marvin-Beckmann (changes requested):
The try_from trait is not needed anymore, as no error can occur.
THere is still a note in the doccomments, that the modulus has to be prime. This needs to be removed, as in the struct description

Marvin-Beckmann (changes requested):
the struct description still contains primeness as an information

Marvin-Beckmann (approved):

PR #287, opened 2023-05-26
Mul for MatZq and MatZ

Marcel583:
**Description**

This PR implements multiplication for MatZq and MatZ.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_zq/arithmetic/mul.rs
+++ b/src/integer_mod_q/mat_zq/arithmetic/mul.rs
@@ -16,7 +16,8 @@ use crate::macros::arithmetics::{
     arithmetic_trait_reverse,
 };
 use crate::traits::{GetNumColumns, GetNumRows};
-use flint_sys::fmpz_mod_mat::fmpz_mod_mat_mul;
+use flint_sys::fmpz_mat::fmpz_mat_mul;
+use flint_sys::fmpz_mod_mat::{_fmpz_mod_mat_reduce, fmpz_mod_mat_mul};
 use std::ops::Mul;
 
 impl Mul for &MatZq {
@@ -53,6 +54,9 @@ impl Mul for &MatZq {
     }
 }
 
+arithmetic_trait_borrowed_to_owned!(Mul, mul, MatZq, MatZq, MatZq);
+arithmetic_trait_mixed_borrowed_owned!(Mul, mul, MatZq, MatZq, MatZq);
+
 impl Mul<&MatZ> for &MatZq {
     type Output = MatZq;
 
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/mat_zq/arithmetic/mul.rs
```diff
@@ -51,6 +53,55 @@ impl Mul for &MatZq {
...
+    fn mul(self, other: &MatZ) -> Self::Output {
+        if self.get_num_columns() != other.get_num_rows() {
+            panic!("Tried to multiply matrices with mismatching matrix dimensions.");
+        }
+
+        let other_matzq = MatZq::from((other, &self.get_mod()));
+        let mut new =
+            MatZq::new(self.get_num_rows(), other.get_num_columns(), self.get_mod()).unwrap();
+        unsafe { fmpz_mod_mat_mul(&mut new.matrix, &self.matrix, &other_matzq.matrix) };
+        new
```
Marvin-Beckmann, line 94: This way we do not have to copy the MatZ. Consider adding reduce by a separate function (as I proposed in a different Kanboard task), but if not, I would be fine with my proposal as well.
```suggestion
        let mut new =
            MatZq::new(self.get_num_rows(), other.get_num_columns(), self.get_mod()).unwrap();
        unsafe {
            fmpz_mat_mul(&mut new.matrix.mat[0], &self.matrix.mat[0], &other.matrix);
            _fmpz_mod_mat_reduce(&mut new.matrix)
        }
        new
```

src/integer_mod_q/mat_zq/arithmetic/mul.rs
```diff
@@ -51,6 +53,55 @@ impl Mul for &MatZq {
...
+    ///
+    /// let c = &a * &b;
+    /// let d = a * b;
+    /// let e = &c * d;
+    /// let f = c * &e;
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Panics if the dimensions of `self` and `other` do not match for multiplication.
+    fn mul(self, other: &MatZ) -> Self::Output {
```
Marvin-Beckmann, line 85: Previously we always added mul_safe as a function that does not panic. Is there any reason you did it differently?

Marvin-Beckmann (changes requested):
Generally looks good, just some changes for efficiency and a general question

src/integer_mod_q/mat_zq/arithmetic/mul.rs
Marcel583: We would need mul_safe for both directions (or a macro for doing so). I thought that this might be unnecessary. But if you think we could need it, I can add it as well.

Marvin-Beckmann (approved):

PR #289, opened 2023-05-26
Refactor from for z

MoogSven:
**Description**
1. Rework the `From` trait for `Z` to use generics.
As a result, the documentation is shorter and references of rust integers are now also supported.

2. `Zq` can no longer be converted into `Z` with the from trait.
The reason for this is, that it is unclear which representation will be taken.
e.g. 3 mod 5 into 3, 8, 13, or ...?
For this reason, we prefer a dedicated function with an appropriate documentation (`get_value`).
As discussed during lunch.
     => Remove uses of `Z::from(Zq)` and `Z::from(&Zq)` (Sampling and distance)
    
3. Also fix doc comment for `Zq::distance`.
**Testing**
Some test cases have been joined (Modulus test into tests_from_int since it uses that Generic implementation)
Added test cases for the reference types, otherwise relied on existing test cases.

**Checklist:**

### DIFF
--- a/src/integer/z/fmpz_helpers.rs
+++ b/src/integer/z/fmpz_helpers.rs
@@ -172,7 +172,7 @@ unsafe impl AsInteger for fmpz {
 
     /// Documentation at [`AsInteger::get_fmpz_ref`]
     fn get_fmpz_ref(&self) -> Option<&fmpz> {
-        Some(&self)
+        Some(self)
     }
 }
 
--- a/src/integer/z/fmpz_helpers.rs
+++ b/src/integer/z/fmpz_helpers.rs
@@ -165,9 +165,7 @@ unsafe impl AsInteger for &fmpz {
 unsafe impl AsInteger for fmpz {
     /// Documentation at [`AsInteger::into_fmpz`]
     unsafe fn into_fmpz(self) -> fmpz {
-        let mut value = fmpz(0);
-        fmpz_set(&mut value, &self);
-        value
+        (&self).into_fmpz()
     }
 
     /// Documentation at [`AsInteger::get_fmpz_ref`]
--- a/src/integer/z/fmpz_helpers.rs
+++ b/src/integer/z/fmpz_helpers.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Niklas Siemer
+// Copyright © 2023 Niklas Siemer, Sven Moog
 //
 // This file is part of qFALL-math.
 //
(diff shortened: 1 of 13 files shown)


src/integer/z/fmpz_helpers.rs
```diff
@@ -148,6 +148,34 @@ unsafe impl AsInteger for &Z {
...
+        Some(self)
+    }
+}
+
+unsafe impl AsInteger for fmpz {
+    /// Documentation at [`AsInteger::into_fmpz`]
+    unsafe fn into_fmpz(self) -> fmpz {
+        let mut value = fmpz(0);
+        fmpz_set(&mut value, &self);
+        value
```
jnsiemer, line 170:
```suggestion
        (&self).into_fmpz()
```

src/integer/z/from.rs
```diff
@@ -265,33 +172,38 @@ impl Z {
...
-        Z::from_modulus(&value)
+implement_empty_trait_owned_ref!(IntoZ for Modulus u8 u16 u32 u64 i8 i16 i32 i64);
+
+impl<Integer: AsInteger + IntoZ> From<Integer> for Z {
+    /// Convert an integer to [`Z`].
+    ///
+    /// # Parameters:
+    /// `value` must be a rust integer, [`Modulus`], or a reference of these types.
+    ///
+    /// # Examples:
```
jnsiemer, line 188:
```suggestion
    /// # Parameters:
    /// - `value` must be a rust integer, [`Modulus`], or a reference of these types.
    ///
    /// Returns something!
    ///
    /// # Examples:
```

src/integer_mod_q/modulus/fmpz_helpers.rs
```diff
@@ -0,0 +1,41 @@
...
+use super::Modulus;
+use crate::traits::AsInteger;
+use flint_sys::fmpz::{fmpz, fmpz_init_set};
+
+unsafe impl AsInteger for Modulus {
+    /// Documentation at [`AsInteger::into_fmpz`]
+    unsafe fn into_fmpz(self) -> fmpz {
+        let mut out = fmpz(0);
+        fmpz_init_set(&mut out, &self.modulus.n[0]);
+        out
```
jnsiemer, line 20:
```suggestion
        (&self).into_fmpz()
```

src/integer_mod_q/z_q/distance.rs
```diff
@@ -98,6 +98,17 @@ impl Distance<&Zq> for Zq {
     }
 }
 
+impl Distance<Zq> for Zq {
+    type Output = Z;
+
+    // This reference links to the correct distance implementation
+    // just because the correct one is the first in this file.
+    /// Just calls [`Zq::distance<&Zq>`].
+    fn distance(&self, other: Zq) -> Self::Output {
```
jnsiemer, line 107: Missing doc comment (referrer would be sufficient)

jnsiemer (approved):
Please address the remaining issues before merging.
Otherwise, great work!
A short explanation of the IntoZ trait for the entire group would be nice as its only needed because Zq also implements `AsInteger` and that's not too obvious.

src/integer_mod_q/z_q/distance.rs
MoogSven: The last comment is already a doc comment.

PR #313, opened 2023-06-13
Z `sample_prime_uniform`

jnsiemer:
**Description**
This PR implements...
- [x] sample_prime_uniform

for `Z`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/sample/uniform.rs
+++ b/src/integer/z/sample/uniform.rs
@@ -39,17 +39,81 @@ impl Z {
     /// - Returns a [`MathError`] of type [`InvalidInterval`](MathError::InvalidInterval)
     /// if the given `upper_bound` isn't at least bigger than `lower_bound + 1`,
     /// i.e. the interval size is at most `1`.
-    pub fn sample_uniform<T1, T2>(lower_bound: &T1, upper_bound: &T2) -> Result<Self, MathError>
-    where
-        T1: Into<Z> + Clone,
-        T2: Into<Z> + Clone,
-    {
-        let lower_bound: Z = lower_bound.clone().into();
-        let upper_bound: Z = upper_bound.clone().into();
+    pub fn sample_uniform(
+        lower_bound: impl Into<Z>,
+        upper_bound: impl Into<Z>,
+    ) -> Result<Self, MathError> {
+        let lower_bound: Z = lower_bound.into();
+        let upper_bound: Z = upper_bound.into();
 
         let interval_size = &upper_bound - &lower_bound;
         let sample = sample_uniform_rejection(&interval_size)?;
         Ok(&lower_bound + sample)
+    }
+
+    /// Chooses a prime [`Z`] instance uniformly at random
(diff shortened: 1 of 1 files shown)


src/integer/z/sample/uniform.rs
```diff
@@ -39,18 +39,82 @@ impl Z {
...
+    ) -> Result<Self, MathError> {
+        let lower_bound: Z = lower_bound.into();
+        let upper_bound: Z = upper_bound.into();
 
         let interval_size = &upper_bound - &lower_bound;
         let sample = sample_uniform_rejection(&interval_size)?;
         Ok(&lower_bound + sample)
     }
+
+    /// Chooses a prime [`Z`] instance uniformly at random
```
Marvin-Beckmann, line 54: Hmm, what do you think, should we return a dedicated error, if the range does not have any prime?
Can we even check something like that?

src/integer/z/sample/uniform.rs
```diff
@@ -142,3 +206,100 @@ mod test_sample_uniform {
...
+        let res_1 = Z::sample_prime_uniform(&lb_1, &upper_bound);
+        let res_2 = Z::sample_prime_uniform(&lb_2, &upper_bound);
+
+        assert!(res_0.is_err());
+        assert!(res_1.is_err());
+        assert!(res_2.is_err());
+    }
+
+    /// Checks whether `sample_prime_uniform` is available for types
+    /// implementing Into<Z>, i.e. u8, u16, u32, u64, i8, ...
```
Marvin-Beckmann, line 288:
```suggestion
    /// implementing Into<Z>, i.e. [u8], [u16], [u32], [u64], [i8], ...
```

src/integer/z/sample/uniform.rs
```diff
@@ -100,7 +164,7 @@ mod test_sample_uniform {
     }
 
     /// Checks whether `sample_uniform` is available for all types
-    /// implementing Into<Z> + Clone, i.e. u8, u16, u32, u64, i8, ...
+    /// implementing Into<Z>, i.e. u8, u16, u32, u64, i8, ...
```
Marvin-Beckmann, line 167:
```suggestion
    /// implementing Into<Z>, i.e. [u8], [u16], [u32], [u64], [i8], ...
```

src/integer/z/sample/uniform.rs
```diff
@@ -142,3 +206,100 @@ mod test_sample_uniform {
...
+mod test_sample_prime_uniform {
+    use crate::{integer::Z, integer_mod_q::Modulus};
+    use std::str::FromStr;
+
+    /// Checks whether `sample_prime_uniform` outputs a prime sample every time.
+    #[test]
+    fn is_prime() {
+        for _ in 0..8 {
+            let sample = Z::sample_prime_uniform(20, i64::MAX).unwrap();
+            assert!(sample.is_prime());
```
Marvin-Beckmann, line 220: maybe add a comment in the assert, that it can potentially fail, even though only with very small probability

Marvin-Beckmann (approved):

PR #318, opened 2023-06-15
Unnecessary Results

jnsiemer:
**Description**
This PR removes the `Result`s from Matrix `new` and `identity` functions. Errors are replaced by suitable `panic` texts.

**Checklist:**


MoogSven (approved):
Nice work. Looking forward to use less unwraps.

PR #322, opened 2023-06-19
Rework from for zq

MoogSven:
**Description**

This PR implements...
- [x] From Trait for Zq
- [x] Remove TryFrom trait for Zq
- [x] Remove `Zq::try_from_int_int` and `Zq::try_from_z_z`
Reasoning:
1. We can use the from trait now that does not return a Result.
2. If we would like to have a function that returns a Result, the naming would have to be changed anyway (maybe to something like `try_from_int_modulus`) and it should use generics.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/z_q/from.rs
+++ b/src/integer_mod_q/z_q/from.rs
@@ -145,12 +145,10 @@ impl Zq {
     }
 }
 
-impl<IntegerValue: Into<Z>, IntegerModulus: Into<Z>> TryFrom<(IntegerValue, IntegerModulus)>
+impl<IntegerValue: Into<Z>, IntegerModulus: Into<Modulus>> From<(IntegerValue, IntegerModulus)>
     for Zq
 {
-    type Error = MathError;
-    /// Implements the [`TryFrom`] trait. It is used to create [`Zq`] from a tuple
-    /// with two values that can be converted into [`Z`].
+    /// Create [`Zq`] from a tuple with the values and the modulus.
     ///
     /// Parameters:
     /// - `value_modulus_tuple` is a tuple of integers `(value, modulus)`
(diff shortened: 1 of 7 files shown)


src/integer_mod_q/z_q/from.rs
```diff
@@ -100,57 +54,12 @@ impl Zq {
...
 }
 
-impl<IntegerValue: Into<Z>, IntegerModulus: Into<Z>> TryFrom<(IntegerValue, IntegerModulus)>
+impl<IntegerValue: Into<Z>, IntegerModulus: Into<Modulus>> From<(IntegerValue, IntegerModulus)>
     for Zq
 {
-    type Error = MathError;
-    /// Implements the [`TryFrom`] trait. It is used to create [`Zq`] from a tuple
-    /// with two values that can be converted into [`Z`].
+    /// Create [`Zq`] from a tuple with the values and the modulus.
```
Marcel583, line 62:
```suggestion
    /// Create [`Zq`] from a tuple with the integer and the modulus.
```

src/integer_mod_q/z_q/from.rs
```diff
@@ -272,112 +176,126 @@ mod test_from_z_modulus {
...
     }
 
     /// Test with zero modulus (not valid)
     #[test]
+    #[should_panic]
     fn modulus_zero() {
-        let new_zq = Zq::try_from((10, 0));
-
-        assert!(new_zq.is_err());
+        let _1 = Zq::from((10, 0));
```
Marcel583, line 291:
```suggestion
        let _ = Zq::from((10, 0));
```

src/integer_mod_q/z_q/from.rs
```diff
@@ -272,112 +176,126 @@ mod test_from_z_modulus {
...
     }
 
     /// Test with negative modulus (not valid)
     #[test]
+    #[should_panic]
     fn modulus_negative() {
-        let new_zq = Zq::try_from((10, -1));
-
-        assert!(new_zq.is_err());
+        let _1 = Zq::from((10, -1));
```
Marcel583, line 298:
```suggestion
        let _ = Zq::from((10, -1));
```

Marcel583 (approved):

PR #332, opened 2023-07-03
Add PolyOverZq from Zq

MoogSven:
It is a common use case to create a Polynomial with just the constant coefficient. These can be used for calculations with other polynomials.

**Description**

This PR implements...
- [x] Add From<&Zq> for Zq
- [x] From<Into<Zq>> for PoyOverZq

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/poly_over_zq/from.rs
+++ b/src/integer_mod_q/poly_over_zq/from.rs
@@ -172,7 +172,7 @@ impl FromStr for PolyOverZq {
     }
 }
 
-impl<IntoZq: Into<Zq>> From<IntoZq> for PolyOverZq {
+impl<IntegerModQ: Into<Zq>> From<IntegerModQ> for PolyOverZq {
     /// Create a constant [`PolyOverZq`] with a specified constant.
     ///
     /// # Parameters:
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/poly_over_zq/from.rs
```diff
@@ -172,6 +172,105 @@ impl FromStr for PolyOverZq {
     }
 }
 
+impl<IntoZq: Into<Zq>> From<IntoZq> for PolyOverZq {
```
Marvin-Beckmann, line 175: IntoZq is kind of irritating to me and does not really fit to our naming (Integer, Rational, etc)

My suggestions:
- ResidueClass
- CongruenceClass
- IntegerModQ

src/integer_mod_q/poly_over_zq/from.rs
```diff
@@ -172,6 +172,105 @@ impl FromStr for PolyOverZq {
...
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the provided value can not be converted into a [`Zq`].
+    ///   For example, if the modulus is not larger than one.
+    fn from(value: IntoZq) -> Self {
+        let value: Zq = value.into();
+        let poly_z = PolyOverZ::from(value.value);
+
+        PolyOverZq::from((&poly_z, &value.modulus))
```
Marvin-Beckmann, line 204: provide ownership for the poly_z if it is later changed to a more efficient variant

Marvin-Beckmann (approved):

src/integer_mod_q/poly_over_zq/from.rs
MoogSven: Selected `IntegerModQ`

src/integer_mod_q/poly_over_zq/from.rs
MoogSven: PolyOverZq::from is not yet implemented for the ownership variant.
Therefore, I left it how it is at the moment.

PR #333, opened 2023-07-03
Division for MatZ by Z

Marcel583:
**Description**

This PR implements division for MatZ by Z.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_z/arithmetic/div.rs
+++ b/src/integer/mat_z/arithmetic/div.rs
@@ -1,0 +1,153 @@
+// Copyright © 2023 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Implementation of scalar multiplication for [`MatZ`] matrices.
+
+use super::super::MatZ;
+use crate::integer::Z;
+use crate::macros::arithmetics::{
+    arithmetic_trait_borrowed_to_owned, arithmetic_trait_mixed_borrowed_owned,
+};
+use crate::traits::{GetNumColumns, GetNumRows};
+use flint_sys::fmpz_mat::fmpz_mat_scalar_divexact_fmpz;
+use std::ops::Div;
+
+impl Div<&Z> for &MatZ {
+    type Output = MatZ;
+    /// Implements division for a [`MatZ`] matrix by a [`Z`] integer.
+    ///
+    /// Parameters:
+    /// - `divisor`: specifies the divisor by which the matrix is divided
+    ///
+    /// Returns the product of `self` and `other` as a [`MatZ`].
+    ///
+    /// Note that for all values in [`MatZ`] it is assumed that they are divisible by the divisor.
+    /// Otherwise, arbitrary values can appear.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::{MatZ, Z};
+    /// use std::str::FromStr;
+    ///
+    /// let mat1 = MatZ::from_str("[[3,6],[9,27]]").unwrap();
+    /// let integer = Z::from(3);
+    ///
+    /// let mat2 = &mat1 / &integer;
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the divisor is `0`.
+    fn div(self, scalar: &Z) -> Self::Output {
+        if scalar == &Z::ZERO {
+            panic!("DivisionByZero: tried to divide {} by a Z of value 0", self);
+        }
+
+        let mut out = MatZ::new(self.get_num_rows(), self.get_num_columns());
+        unsafe {
+            fmpz_mat_scalar_divexact_fmpz(&mut out.matrix, &self.matrix, &scalar.value);
+        }
+        out
+    }
+}
+
+arithmetic_trait_borrowed_to_owned!(Div, div, MatZ, Z, MatZ);
+arithmetic_trait_mixed_borrowed_owned!(Div, div, MatZ, Z, MatZ);
+
+#[cfg(test)]
+mod test_div {
+    use super::MatZ;
+    use crate::integer::Z;
+    use std::str::FromStr;
+
+    /// Checks if matrix division works fine for both borrowed
+    #[test]
+    fn borrowed_correctness() {
+        let mat1 = MatZ::from_str("[[6,3],[3,6]]").unwrap();
+        let mat3 = MatZ::from_str("[[2,1],[1,2]]").unwrap();
+        let integer = Z::from(3);
+
+        let mat1 = &mat1 / &integer;
+
+        assert_eq!(mat3, mat1);
+    }
+
+    /// Checks if matrix division works fine for both owned
+    #[test]
+    fn owned_correctness() {
+        let mat1 = MatZ::from_str("[[6,3],[3,6]]").unwrap();
+        let mat3 = MatZ::from_str("[[2,1],[1,2]]").unwrap();
+        let integer = Z::from(3);
+
+        let mat1 = mat1 / integer;
+
+        assert_eq!(mat3, mat1);
+    }
+
+    /// Checks if matrix division works fine for half owned/borrowed
+    #[test]
+    fn half_correctness() {
+        let mat1 = MatZ::from_str("[[6,3],[3,6]]").unwrap();
+        let mat2 = mat1.clone();
+        let mat3 = MatZ::from_str("[[2,1],[1,2]]").unwrap();
+        let integer1 = Z::from(3);
+
+        let mat1 = mat1 / &integer1;
+        let mat2 = &mat2 / integer1;
+
+        assert_eq!(mat3, mat1);
+        assert_eq!(mat3, mat2);
+    }
+
+    /// Checks if matrix division works fine for matrices of different dimensions
+    #[test]
+    fn different_dimensions_correctness() {
+        let mat1 = MatZ::from_str("[[3],[0],[12]]").unwrap();
+        let mat2 = MatZ::from_str("[[6,15,18],[3,9,3]]").unwrap();
+        let mat3 = MatZ::from_str("[[1],[0],[4]]").unwrap();
+        let mat4 = MatZ::from_str("[[2,5,6],[1,3,1]]").unwrap();
+        let integer = Z::from(3);
+
+        assert_eq!(mat3, mat1 / &integer);
+        assert_eq!(mat4, mat2 / integer);
+    }
+
+    /// Checks if matrix division works fine for large values
+    #[test]
+    fn large_entries() {
+        let mat1 = MatZ::from_str(&format!("[[6],[{}],[2]]", i64::MAX / 3)).unwrap();
+        let mat2 = MatZ::from_str(&format!("[[{}]]", i64::MAX)).unwrap();
+        let mat3 = MatZ::from_str(&format!("[[3],[{}],[1]]", (i64::MAX / 6))).unwrap();
+        let mat4 = MatZ::from_str("[[1]]").unwrap();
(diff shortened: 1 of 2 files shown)


src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,153 @@
...
+arithmetic_trait_borrowed_to_owned!(Div, div, MatZ, Z, MatZ);
+arithmetic_trait_mixed_borrowed_owned!(Div, div, MatZ, Z, MatZ);
+
+#[cfg(test)]
+mod test_div {
+    use super::MatZ;
+    use crate::integer::Z;
+    use std::str::FromStr;
+
+    /// Checks if matrix division works fine for both borrowed
```
MoogSven, line 67: Finish sentences with a dot

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,153 @@
...
+    #[test]
+    fn borrowed_correctness() {
+        let mat1 = MatZ::from_str("[[6,3],[3,6]]").unwrap();
+        let mat3 = MatZ::from_str("[[2,1],[1,2]]").unwrap();
```
MoogSven, line 71: mat3 is not required for calculation => move below above the assert

Similar for tests below.

MoogSven (changes requested):

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,153 @@
...
+use super::super::MatZ;
+use crate::integer::Z;
+use crate::macros::arithmetics::{
+    arithmetic_trait_borrowed_to_owned, arithmetic_trait_mixed_borrowed_owned,
+};
+use crate::traits::{GetNumColumns, GetNumRows};
+use flint_sys::fmpz_mat::fmpz_mat_scalar_divexact_fmpz;
+use std::ops::Div;
+
+impl Div<&Z> for &MatZ {
```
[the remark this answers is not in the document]
MoogSven: Not using Generics, because that would be too much effort at the moment (Requires new macro).
Might be switched to generic when other functions are also updated.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+    type Output = MatQ;
+    /// Implements division for a [`MatZ`] matrix by a [`Z`] integer.
+    ///
+    /// Parameters:
+    /// - `divisor`: specifies the divisor by which the matrix is divided
+    ///
+    /// Returns the product of `self` and `other` as a [`MatQ`] or panics if
+    /// the divisor is 0.
```
MoogSven, line 30:
```suggestion
    /// Returns the quotient of `self` divided by `other` as a [`MatQ`]
    /// or panics if the divisor is 0.
```

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::{MatZ, Z};
+    /// use std::str::FromStr;
+    ///
+    /// let mut mat1 = MatZ::from_str("[[3,6],[9,27]]").unwrap();
+    /// let integer = Z::from(3);
+    ///
+    /// let mat2 = unsafe{ mat1.div_exact(&integer)};
+    /// ```
```
MoogSven, line 84: 1. You can use 3 directly, you don't need to create a Z first.
2. Please add an assert so that one can see how the result looks like.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// use qfall_math::integer::{MatZ, Z};
+    /// use std::str::FromStr;
+    ///
+    /// let mat1 = MatZ::from_str("[[3,6],[9,27]]").unwrap();
+    /// let integer = Z::from(3);
+    ///
+    /// let mat2 = unsafe{&mat1.div_exact_ref(&integer)};
+    /// ```
+    #[allow(clippy::missing_safety_doc)]
```
MoogSven, line 116:
```suggestion
```
Why is this here? You have a # Safety section.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+        self
+    }
+
+    /// Implements division for a [`MatZ`] matrix by a [`Z`] integer.
+    ///
+    /// Parameters:
+    /// - `divisor`: specifies the divisor by which the matrix is divided
+    ///
+    /// Returns the product of `self` and `other` as a [`MatZ`] or panics if
+    /// the divisor is 0.
```
MoogSven, line 99: See above comment for owned version.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        out
+    }
+}
+
+#[cfg(test)]
+mod test_div_exact {
+    use super::*;
+    use std::str::FromStr;
+
```
MoogSven, line 131: 1. Add a test to showcase the different types with which the function can be called with. e.g. Z,  Modulus, rust integers (owned and borrowed).
Then, you can also replace borrowed_ , owned_ , and mixed_correctness with one test e.g. small_numbers.

2. Also, you can use the rust integers directly in the tests below to make them ore simple instead of creating a Z first. If you stick with the variable for the divisor, you should also call it "divisor" and not just "integer".

3. Add (one or more) tests where negative matrix entries are divided.

4. Test with negative divisor.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+        let integer1 = Z::from(3);
+        let integer2 = Z::from(3);
+
+        let mat1 = unsafe { mat1.div_exact_ref(integer1) };
+        let mat2 = unsafe { mat2.div_exact(integer2) };
+
+        let mat3 = MatZ::from_str("[[2,1],[1,2]]").unwrap();
+
+        assert_eq!(mat3, mat1);
+        assert_eq!(mat3, mat2);
```
MoogSven, line 162: Move assert logic together, rename `mat3` to a more meaningful `mat_cmp`.
Also applies to following tests.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+
+        let mut out = MatZ::new(self.get_num_rows(), self.get_num_columns());
+        unsafe {
+            fmpz_mat_scalar_divexact_fmpz(&mut out.matrix, &self.matrix, &divisor.value);
+        }
+        out
+    }
+}
```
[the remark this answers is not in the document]
Marcel583: This is not my current implementation, but an outdated one

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+        unsafe {
+            fmpq_mat_set_fmpz_mat_div_fmpz(&mut out.matrix, &self.matrix, &divisor.value);
+        }
+        out
+    }
+}
+
+arithmetic_trait_borrowed_to_owned!(Div, div, MatZ, Z, MatQ);
+arithmetic_trait_mixed_borrowed_owned!(Div, div, MatZ, Z, MatQ);
+
```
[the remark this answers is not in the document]
Marcel583: We do not need it and to implement it we would need another macro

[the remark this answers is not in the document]
MoogSven: ok...

src/integer/mat_z/arithmetic/div.rs
MoogSven: Some clarifications of what i meant with this:
> Add a test to showcase the different types with which the function can be called with. e.g. Z, Modulus, rust integers (owned and borrowed).
Then, you can also replace borrowed_ , owned_ , and mixed_correctness with one test e.g. small_numbers.

I expected you to create one test to test all the different type combinations. We already have 30 tests like that called `availability`. This way the code should become much shorter because you don't have to initialize mat1, mat2, mat_cmp all the time.

Additionally, you don't have to create a `mat2` if you don't shadow `mat1` and just clone when you call the function or by first using `div_exact_ref` and then `div_exact`.
e.g. for availability:
``` rust
let mat = MatZ::from_str(...).unwrap();

// div_exact_ref
let _ = unsafe{ mat.div_exact_ref(3i64)};
let _ = unsafe{ mat.div_exact_ref(3u32)};
let _ = unsafe{ mat.div_exact_ref(Z::from(3))};
[... Also for the borrowed versions and maybe some more different rust integer types]

/// div_exact
let _ = unsafe{ mat.clone().div_exact(3i64) };
let _ = unsafe{ mat.clone().div_exact(3iu32) };
[... Also for the borrowed versions and maybe some more different rust integer types]
```

For the other tests:
```rust
let mat = [...]

let mat_divided_1 = unsafe{ mat.div_exact_ref(3)};
let mat_divided_2 = unsafe{ mat.div_exact(3) };

let mat_cmp = [...]
assert_eq!(mat_divided_1, mat_cmp);
assert_eq!(mat_divided_2, mat_cmp);
```

> 2. Also, you can use the rust integers directly in the tests below to make them ore simple instead of creating a Z first. If you stick with the variable for the divisor, you should also call it "divisor" and not just "integer".

Not sure why you did not do this.

> 3. Add (one or more) tests where negative matrix entries are divided.
> 4. Test with negative divisor.

Do the negative test in a dedicated test function. `types_correctness` and `different_dimensions_correctness` do not sound like checking negative divisors/values and the comment also do not mention it.
Furthermore, also test with large negative divisors and large negative values.
You can combine 3. and 4. within the same test.

Additionally, you could consider removing the `different_dimensions_correctness` test and just use different dimensions throughout the other tests.

src/integer/mat_z/arithmetic/div.rs
MoogSven: > Please add an assert so that one can see how the result looks like.

This also applies to the other doc tests.

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Checks if matrix division works with not dividable entries.
+    #[test]
+    fn not_dividable_error() {
+        let mat1 = MatZ::from_str("[[6,2],[3,10]]").unwrap();
+        let integer = Z::from(3);
+
+        let _mat1 = &mat1 / &integer;
+    }
```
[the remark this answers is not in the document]
MoogSven: Why does this test have error in it's name?
It is a regular division.
Where is the assert?

src/integer/mat_z/arithmetic/div.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        assert_eq!(mat3, unsafe { mat1.div_exact_ref(&integer1) });
+        assert_eq!(mat4, unsafe { mat2.div_exact_ref(&integer2) });
+        assert_eq!(mat3, unsafe { mat1.div_exact(&integer1) });
+        assert_eq!(mat4, unsafe { mat2.div_exact(&integer2) });
+    }
+}
+
+#[cfg(test)]
+mod test_div {
```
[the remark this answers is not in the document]
MoogSven: Also the case for the updated suggestions.

MoogSven (changes requested):
Check previous suggestions, I commented on those with clearifications.

src/integer/mat_z/arithmetic/div.rs
Marcel583: I would call it integer because we call the functionality "division by integer"

src/integer/mat_z/arithmetic/div.rs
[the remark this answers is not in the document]
Marcel583: All suggestions are applied here

MoogSven (approved):
I like the new tests :+1:.
Added some suggestions you should look over before merging.

PR #335, opened 2023-07-03
Inverse for MatZq with prime modulus

Marcel583:
**Description**

This PR implements inverse for MatZq with prime modulus.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_zq/inverse.rs
+++ b/src/integer_mod_q/mat_zq/inverse.rs
@@ -1,0 +1,227 @@
+// Copyright © 2023 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains the implementation of the `inverse` function.
+//a commutative ring, is invertible iff its determinant is invertible
+use super::MatZq;
+use crate::{
+    integer::{MatZ, Z},
+    integer_mod_q::Zq,
+    traits::{Concatenate, Gcd, GetEntry, GetNumColumns, GetNumRows, Pow},
+};
+
+impl MatZq {
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant unequal to zero) and `None` otherwise.
+    ///
+    /// Note that the modulus is assumed to be prime.
+    /// If it is not, it can happen that the function panics.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    /// use qfall_math::traits::*;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZ::from_str("[[1,2],[3,4]]").unwrap();
+    /// let matrix_invert = matrix.inverse().unwrap();
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the inverse of an entry can not be computed because the modulus is not prime
+    pub fn inverse_prime(&self) -> Option<MatZq> {
+        // check if determinant is coprime to modulus
+        let det = MatZ::from(self).det().ok()?;
+        if det.gcd(Z::from(self.get_mod())) != Z::ONE {
+            None
+        } else {
+            let matrix_identity = self
+                .concat_horizontal(&MatZq::identity(
+                    self.get_num_rows(),
+                    self.get_num_columns(),
+                    self.get_mod(),
+                ))
+                .unwrap();
+
+            let identity_inverse = matrix_identity.gaussian_elimination_prime();
+
+            let mut inverse =
+                MatZq::new(self.get_num_rows(), self.get_num_columns(), self.get_mod());
+            for i in 0..self.get_num_columns() {
+                inverse
+                    .set_column(i, &identity_inverse, self.get_num_columns() + i)
+                    .unwrap();
+            }
+            Some(inverse)
+        }
+    }
+
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant unequal to zero) and `None` otherwise.
+    ///
+    /// Note that the modulus is assumed to be prime.
+    /// If it is not, it can happen that the function panics.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    /// use qfall_math::traits::*;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZ::from_str("[[1,2],[3,4]]").unwrap();
+    /// let matrix_invert = matrix.inverse().unwrap();
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the inverse of an entry can not be computed because the modulus is not prime
+    pub fn gaussian_elimination_prime(&self) -> MatZq {
+        let mut out = self.clone();
+        let mut row_count = 0;
+        for column_nr in 0..self.get_num_columns() {
+            if row_count >= self.get_num_rows() {
+                break;
+            }
+
+            let mut current_row = -1;
+            let mut entry = Zq::from((1, self.get_mod()));
+            for row_nr in row_count..self.get_num_rows() {
+                entry = out.get_entry(row_nr, column_nr).unwrap();
+                if !Zq::is_zero(&entry) {
+                    current_row = row_nr;
+                    break;
+                }
+            }
+            if current_row == -1 {
+                continue;
+            }
+
+            if let Ok(inv) = entry.pow(-1) {
+                let row = inv * out.get_row(current_row).unwrap();
+                out.set_row(current_row, &row, 0).unwrap();
+
+                // set all other entries in that column to `0` (gaussian elimination)
+                for row_nr_other in (0..self.get_num_rows()).filter(|x| *x != current_row) {
+                    let old_row = out.get_row(row_nr_other).unwrap();
+                    let entry: Z = old_row.get_entry(0, column_nr).unwrap();
+                    let new_row = &old_row - entry * &row;
+                    out.set_row(row_nr_other, &new_row, 0).unwrap();
+                }
+
+                if row_count != current_row {
+                    out.swap_rows(row_count, current_row).unwrap();
+                }
+                row_count += 1;
+            } else {
+                panic!("The modulus {} is not prime", self.get_mod());
+            }
+        }
+
+        out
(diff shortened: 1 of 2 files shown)


src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,228 @@
...
+#[cfg(test)]
+mod test_inverse {
+    use crate::{integer::MatZ, integer_mod_q::MatZq, rational::MatQ};
+    use std::str::FromStr;
+
+    /// Test whether `inverse` correctly calculates an inverse matrix.
+    #[test]
+    fn inverse_works() {
+        let mat1 = MatZq::from_str("[[5,2],[2,1]] mod 7").unwrap();
+        let cmp_inv1 = MatZq::from_str("[[1, 5],[5, 5]] mod 7").unwrap();
```
MoogSven, line 178: Rename mat1 to mat, move the creation of mat_cmp down.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        assert_eq!(cmp_inv1, inv1);
+    }
+
+    /// Check if the multiplication of inverse and matrix result in an identity matrix.
+    #[test]
+    fn inverse_correct() {
+        let mat = MatZ::from_str("[[5,2],[2,1]]").unwrap();
+        let mat_q = MatQ::from(&mat);
+        let cmp = MatQ::from_str("[[1,0],[0,1]]").unwrap();
```
MoogSven, line 190: 1. MatQ::identity
2. Why MatQ?!

MoogSven (changes requested):

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,228 @@
+// Copyright © 2023 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains the implementation of the `inverse` function.
+//a commutative ring, is invertible iff its determinant is invertible
```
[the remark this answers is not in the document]
Marcel583: That was not supposed to be there...

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+    traits::{Concatenate, Gcd, GetNumRows},
+};
+use flint_sys::fmpz_mod_mat::fmpz_mod_mat_rref;
+
+impl MatZq {
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant unequal to zero) and `None` otherwise.
+    ///
+    /// Note that the modulus is assumed to be prime.
+    /// If it is not, it can happen that the function panics.
```
MoogSven, line 23: This way, this function is really annoying to use. Will it panic or not when the modulus is not prime, nobody knows, but you will not be able to recover from it.
Also, I think that it will currently always panic in that case.

Suggestion: just allow prime moduli
Also, I don't think that Option is the right return type then. I think Result would fit better, because we can then insert into the error message why it did not work.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
+    /// let matrix_gauss = matrix.gaussian_elimination_prime();
+    ///
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", matrix_gauss.to_string());
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the modulus is not prime.
+    pub fn gaussian_elimination_prime(self) -> MatZq {
```
MoogSven, line 84: Suggestion:
Return a result instead.  It is more rusty and the simplified syntax is not necessary here, since gaussian_elemination_prime will probably not be written in code that often. It also allows programming patterns where the parameter is first thrown into this function, and if an error occurs into the next.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// use qfall_math::integer_mod_q::MatZq;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
+    /// let matrix_invert = matrix.inverse_prime().unwrap();
+    ///
+    /// let id = &matrix_invert * &matrix;
+    ///
+    /// assert_eq!("[[5, 1],[5, 3]] mod 7", matrix_invert.to_string());
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", id.to_string());
```
MoogSven, line 36:
```suggestion
    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
    ///
    /// let matrix_invert = matrix.inverse_prime().unwrap();
    ///
    /// let id = &matrix_invert * &matrix;
    /// assert_eq!("[[5, 1],[5, 3]] mod 7", matrix_invert.to_string());
    /// assert_eq!("[[1, 0],[0, 1]] mod 7", id.to_string());
```
Make it more obvious where the function is called.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer_mod_q::MatZq;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
+    /// let matrix_gauss = matrix.gaussian_elimination_prime();
+    ///
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", matrix_gauss.to_string());
```
MoogSven, line 79:
```suggestion
    /// use qfall_math::integer_mod_q::MatZq;
    /// use std::str::FromStr;
    ///
    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
    ///
    /// let matrix_gauss = matrix.gaussian_elimination_prime();
    ///
    /// assert_eq!("[[1, 0],[0, 1]] mod 7", matrix_gauss.to_string());
```

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Check if the multiplication of inverse and matrix result in an identity matrix.
+    #[test]
+    fn inverse_correct() {
+        let mat = MatZq::from_str("[[5,2],[2,1]] mod 11").unwrap();
+
+        let inv = mat.inverse_prime().unwrap();
+        let diag = mat * inv;
+
+        let cmp = MatZq::from_str("[[1,0],[0,1]] mod 11").unwrap();
```
MoogSven, line 121: use `MatZq::identity`
this makes it more clear what is going on.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+        assert_eq!(cmp, diag);
+    }
```
MoogSven, line 123: Why are `inverse_works` and `inverse_correct` two different test cases? They test the same matrix.

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// Check if the multiplication of inverse and matrix result in an identity matrix.
+    #[test]
+    fn inverse_correct() {
+        let mat = MatZq::from_str("[[5,2],[2,1]] mod 11").unwrap();
+
+        let inv = mat.inverse_prime().unwrap();
+        let diag = mat * inv;
+
+        let cmp = MatZq::from_str("[[1,0],[0,1]] mod 11").unwrap();
+        assert_eq!(cmp, diag);
```
MoogSven, line 122:
```suggestion
        let inv = mat.inverse_prime().unwrap();

        let diag = mat * inv;
        let cmp = MatZq::from_str("[[1,0],[0,1]] mod 11").unwrap();
        assert_eq!(cmp, diag);
```

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        self
+    }
+}
+
+#[cfg(test)]
+mod test_inverse {
+    use crate::integer_mod_q::MatZq;
+    use std::str::FromStr;
+
```
MoogSven, line 101: Add test for non-prime modulus.

MoogSven (changes requested):

src/integer_mod_q/mat_zq/inverse.rs
Marcel583: It should be an option since we always have an option in inverse functions

src/integer_mod_q/mat_zq/inverse.rs
Marcel583: The first test is for the correct inverse, the second one assures that the inverse multiplied by the matrix results in an identity matrix

src/integer_mod_q/mat_zq/inverse.rs
Marcel583: For consistency I would say that it should stay as a panic

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub fn inverse_prime(&self) -> Option<MatZq> {
+        if !self.get_mod().is_prime() {
+            panic!("The modulus of the matrix is not prime");
+        }
+
+        // Check if the matrix is spare and compute determinant
+        if let Ok(det) = MatZ::from(self).det() {
+            // Check if determinant is coprime to modulus.
+            if det.gcd(self.get_mod()) != Z::ONE {
+                None
```
MoogSven, line 49: When can this happen?
The modulus is prime => gcd should always be one.

src/integer_mod_q/mat_zq/inverse.rs
MoogSven: ok...

src/integer_mod_q/mat_zq/inverse.rs
MoogSven: ok...

PR #336, opened 2023-07-03
Refactor sprint 10

Marvin-Beckmann:
**Description**

- [x] fix warnings in sampling (clippy etc.)
- [x] hide multiple occurrences in documentation 
  - *I only removed the documentation from the macros, I did not write down a collection of all supported types, should I do that?*
- [x] rename generics in function meaningful e.g. T -> Integer, maybe directly use Into<Z> and not indirectly with T.
- [x] Some functions require `&Into<Z>` as parameters. This is often indirect e.g. `T: Into<Z> + Clone` and `&T`. They can probably be converted into `Into<Z>` instead. If the Clone trait is required, it can also be removed then. This also applies for `Zq` and `Q`.
- [x] Fix some spelling mistakes along the way
- [x] Remove macrousage and replace with Into's


MoogSven:
> I only removed the documentation from the macros, I did not write down a collection of all supported types, should I do that?
Yes, that would be nice. You can maybe shorten it a bit by saying [...] and their references. Or collecting i8, i16 etc. as rust integers.

src/integer_mod_q/mat_zq/sample/uniform.rs
```diff
@@ -111,7 +111,7 @@ mod test_sample_uniform {
     }
 
     /// Checks whether `sample_uniform` is available for all types
-    /// implementing Into<Z> + Clone, i.e. u8, u16, u32, u64, i8, ...
+    /// implementing [`Into<Z>`] + [`Clone`], i.e. u8, u16, u32, u64, i8, ...
```
MoogSven, line 114: Clone no longer required (fixed in commit)

MoogSven (commented):
Your changes look good. 

Optionally: 
Remove the functions generated by the `arithmetic_between_types_zq` macro from the doc. 
For example, in Zq we have:
```
Add<&Z>
Add<&Zq>
Add<&Zq>
Add<&i16>
Add<&i32>
Add<&i64>
Add<&i8>
Add<&u16>
Add<&u32>
Add<&u64>
Add<&u8>
```
The same For Mul etc.
Maybe we should just port some of the arithmetic between types macros to generics to resolve this issue. I will make some tests on how feasible that is. Not necessary for this PR.

I will also now also check that you did not miss anything/ too much and probably change those things directly.

MoogSven:
> The same For Mul etc.
Maybe we should just port some of the arithmetic between types macros to generics to resolve this issue. I will make some tests on how feasible that is. Not necessary for this PR.

Results from my tests:
- Generic types can be used e.g. 
```impl<Integer: Into<Z>> Add<Integer> for &Zq {```
- Implementing the same for `Zq` instead of `&Zq` can be done with a macro or a generic.
- Implementing the revers direction (not just `Z + i32`, but also `i32 + Z`) has to be done with a macro.

I created a proof of concept branch where you can check it out
https://github.com/qfall/math/tree/arithmetic_generics_proof_of_concept

**Review by MoogSven — APPROVED** [review]

**Comment by Marvin-Beckmann** [comment]
> > The same For Mul etc.
> > Maybe we should just port some of the arithmetic between types macros to generics to resolve this issue. I will make some tests on how feasible that is. Not necessary for this PR.
> 
> Results from my tests:
> 
> * Generic types can be used e.g.
>   `impl<Integer: Into<Z>> Add<Integer> for &Zq {`
> * Implementing the same for `Zq` instead of `&Zq` can be done with a macro or a generic.
> * Implementing the revers direction (not just `Z + i32`, but also `i32 + Z`) has to be done with a macro.
> 
> I created a proof of concept branch where you can check it out https://github.com/qfall/math/tree/arithmetic_generics_proof_of_concept

I would not do this in this refactoring, I think to change all arithmetics we will need more time.
I would suggest to do this at a later point in time

File: src/integer/mat_z/set.rs [hunk 2]
```diff
@@ -31,7 +30,7 @@ use std::{
     ptr::{null, null_mut},
 };
 
-impl SetEntry<&Z> for MatZ {
+impl<Integer: Into<Z>> SetEntry<Integer> for MatZ {
     /// Sets the value of a specific matrix entry according to a given `value` of type [`Z`].
```
MoogSven, line 34: It is no longer only `Z`

src/integer/mat_poly_over_z/evaluate.rs
```diff
@@ -5,12 +5,11 @@
...
-    macros::for_others::{implement_for_others, implement_for_owned},
     traits::{Evaluate, GetNumColumns, GetNumRows},
 };
 use flint_sys::fmpz_poly_mat::fmpz_poly_mat_evaluate_fmpz;
 
-impl Evaluate<&Z, MatZ> for MatPolyOverZ {
+impl<Integer: Into<Z>> Evaluate<Integer, MatZ> for MatPolyOverZ {
     /// Evaluates a [`MatPolyOverZ`] on a given input of [`Z`] entrywise. Note that the
     /// [`Z`] in this case is only a reference.
     ///
```
MoogSven, line 15: Update documentation that it works on  [Into<Z>]. The comment with Z is only a reference is outdated now.

I would suggest to also update the doc comment to once evaluate directly on `3` and once on `Z::from(3)`.

MoogSven (changes requested):
I like that the documentation is now much cleaner. However, you still need to update the comments on many functions where you switch to generics. Some explicitly mention [`Z`], there you could either change it to "Integer" or "[Into<`Z`>]". And some also explicitly mention that the parameter has to be a reference, which is no longer the case.

I would also like to see updated doc tests where (at least) two types are shown to use the function.

MoogSven (approved):
Really nice changes :+1: 

You missed few instances where generics can also be used. You can find them by searching for `implement_for_owned`. I added them in a commit (review it before merging)
- SetEntry for MatZq
- SetEntry for MatQ
- Evaluate for PolyOverZq
- SetEntry for PolyOverZq

PR #352, opened 2023-07-20
Tensor product for MatPolyOverZ

Marvin-Beckmann:
**Description**

This PR implements
- [x] Tensor product for MatPolyOverZ.
It resues an old implementation of the tensor product for MatZ. For MatZ we were able to use a dedicated method, for MatPolyOverZ, we are not able to -> hence reuse that implementation. The corresponding PR: <https://github.com/qfall/math/pull/160>

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_poly_over_z/tensor.rs
+++ b/src/integer/mat_poly_over_z/tensor.rs
@@ -1,0 +1,257 @@
+// Copyright © 2023 Marvin Beckmann
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains the implementation of the `tensor` product.
+
+use super::MatPolyOverZ;
+use crate::{
+    integer::PolyOverZ,
+    traits::{GetEntry, GetNumColumns, GetNumRows, Tensor},
+};
+use flint_sys::{fmpz_poly::fmpz_poly_mul, fmpz_poly_mat::fmpz_poly_mat_entry};
+
+impl Tensor for MatPolyOverZ {
+    /// Computes the tensor product of `self` with `other`
+    ///
+    /// Parameters:
+    /// - `other`: the value with which the tensor product is computed.
+    ///
+    /// Returns the tensor product of `self` with `other`.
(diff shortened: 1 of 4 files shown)


src/integer/mat_poly_over_z/tensor.rs
```diff
@@ -0,0 +1,257 @@
...
+/// - `row_left`: defines the leftmost row of the set window
+/// - `column_upper`: defines the highest column of the set window
+/// - `scalar`: defines the value with which the part of the tensor product
+/// is calculated
+/// - `matrix`: the matrix with which the scalar is multiplied
+/// before setting the entries in `out`
+///
+/// Implicitly sets the entries of the matrix according to the definition
+/// of the tensor product.
+unsafe fn set_matrix_window_mul(
```
MoogSven, line 80: Please add a `# Safety` section to the documentation describing why it is unsafe, i.e., matrix row_left and column_upper must be in a specific range.
If we have negative or too large `row_left` or `column_upper`, (really) bad stuff can happen. Randomly overwriting other data on the heap, and potentially also on the stack.

Maybe add an assert, however, since the function is private and unsafe that might not be necessary.

src/integer/mat_poly_over_z/tensor.rs
```diff
@@ -0,0 +1,257 @@
...
+        let mat_3 = mat_1.tensor_product(&mat_2);
+
+        assert_eq!(51, mat_3.get_num_rows());
+        assert_eq!(52, mat_3.get_num_columns());
+    }
+
+    /// Ensure that the tensor works correctly with identity
+    #[test]
+    fn identity() {
+        let identity = MatPolyOverZ::from_str("[[1  1, 0],[0, 1  1]]").unwrap();
```
MoogSven, line 130:
```suggestion
        let identity = MatPolyOverZ::identity(2,2);
```

src/integer/mat_poly_over_z/tensor.rs
MoogSven: `# Examples` is also missing, but since it is a private function just used in this file, I am fine without it.

MoogSven (approved):
Looks good, just minor changes required.

src/integer/mat_poly_over_z/tensor.rs
Marvin-Beckmann: left out examples, added security section and asserts

PR #356, opened 2023-07-23
Setter for PolyRingZq

Marcel583:
**Description**

This PR implements setter for PolyRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/poly_over_z/set.rs
+++ b/src/integer/poly_over_z/set.rs
@@ -67,7 +67,7 @@ mod test_set_coeff {
 
     /// Ensure that the negative indices return an error.
     #[test]
-    fn set_min_negative_coeff() {
+    fn set_negative_index() {
         let mut poly = PolyOverZ::from_str("2  1 1").unwrap();
 
         assert!(poly.set_coeff(i64::MIN, 2).is_err());
@@ -81,13 +81,13 @@ mod test_set_coeff {
     fn set_coeff_big() {
         let mut poly = PolyOverZ::from_str("2  1 1").unwrap();
 
-        assert!(poly.set_coeff(i8::MAX, 2).is_ok());
-        assert!(poly.set_coeff(i16::MAX, 2).is_ok());
+        assert!(poly.set_coeff(2, i32::MAX).is_ok());
+        assert!(poly.set_coeff(2, i64::MAX).is_ok());
     }
 
     /// Ensure that the max of [`u8`] and [`u16`] works as an index.
     #[test]
-    fn set_unsigned_coeff() {
+    fn set_index_big() {
         let mut poly = PolyOverZ::from_str("2  1 1").unwrap();
 
         assert!(poly.set_coeff(u8::MAX, 2).is_ok());
--- a/src/integer/poly_over_z/set.rs
+++ b/src/integer/poly_over_z/set.rs
@@ -19,8 +19,6 @@ impl<Integer: Into<Z>> SetCoefficient<Integer> for PolyOverZ {
     /// of roughly 34 GB. If not careful, be prepared that memory problems can occur, if
     /// the index is very high.
     ///
-    /// All entries which are not directly addressed are automatically treated as zero.
-    ///
     /// Parameters:
     /// - `index`: the index of the coefficient to set (has to be positive)
     /// - `value`: the new value the index should have
@@ -67,7 +65,7 @@ mod test_set_coeff {
     };
     use std::str::FromStr;
 
-    /// Ensure that the negative indices return an error
+    /// Ensure that the negative indices return an error.
     #[test]
     fn set_min_negative_coeff() {
         let mut poly = PolyOverZ::from_str("2  1 1").unwrap();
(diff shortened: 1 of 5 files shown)


src/integer_mod_q/polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,171 @@
...
+use flint_sys::fmpz_poly::fmpz_poly_set_coeff_fmpz;
+use std::fmt::Display;
+
+impl<Integer: Into<Z>> SetCoefficient<Integer> for PolynomialRingZq {
+    /// Sets the coefficient of a [`PolynomialRingZq`] element.
+    /// We advise to use small coefficients, since already 2^32 coefficients take space
+    /// of roughly 34 GB. If not careful, be prepared that memory problems can occur, if
+    /// the index is very high.
+    ///
+    /// All entries which are not directly addressed are automatically treated as zero.
```
MoogSven, line 22: > All entries which are not directly addressed are automatically treated as zero.

I am aware that this sentence is also part of the other set functions. However, I think it can be wrongly interpreted.
For example as: This function addresses only directly addresses one entry (coefficient would be a better name here in my opinion). Therefore, all other coefficients are set to zero.

It is more like: "If the `index` is larger than the degree of the polynomial, the polynomial is expanded with zero coefficients." However, I think that this is behavior that should be expected. 

My suggestions: remove this sentence and optionally add something like "can be larger than the degree of the polynomial" to the parameter description of `index`.

src/integer_mod_q/polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        poly_ring.set_coeff(1, 3u64).unwrap();
+        poly_ring.set_coeff(1, 3u32).unwrap();
+        poly_ring.set_coeff(1, 3u16).unwrap();
+        poly_ring.set_coeff(1, 3u8).unwrap();
+        poly_ring.set_coeff(1, 3i64).unwrap();
+        poly_ring.set_coeff(1, 3i32).unwrap();
+        poly_ring.set_coeff(1, 3i16).unwrap();
+        poly_ring.set_coeff(1, 3i8).unwrap();
+        poly_ring.set_coeff(1, Z::from(3)).unwrap();
```
MoogSven, line 90:
```suggestion
        poly_ring.set_coeff(1, Z::from(3)).unwrap();
        poly_ring.set_coeff(1, &Z::from(3)).unwrap();
```

src/integer_mod_q/polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Ensure that the max of [`u8`] and [`u16`] works as a coefficient.
+    #[test]
+    fn set_unsigned_coeff() {
+        let modulus = ModulusPolynomialRingZq::from_str("4  1 0 0 1 mod 17").unwrap();
+        let poly = PolyOverZ::from_str("3  0 1 1").unwrap();
+        let mut poly_ring = PolynomialRingZq::from((&poly, &modulus));
+
+        assert!(poly_ring.set_coeff(u8::MAX, 2).is_ok());
+        assert!(poly_ring.set_coeff(u16::MAX, 2).is_ok());
+    }
```
MoogSven, line 113: I think that you can remove `set_max_coeff`. It is basically the same test as `set_unsigned_coeff`, but with a smaller index.

Additionally, the wording should be clearified to more clearly differentiate between `index` and `value`. I thought of value when I read the word coefficient, but the u16::MAX th coefficient is set.

Furthermore, I think you can simplify the test by setting the poly to `PolyOverZ::default()`.

src/integer_mod_q/polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    /// Ensure that the negative indices return an error.
+    #[test]
+    fn set_min_negative_coeff() {
+        let modulus = ModulusPolynomialRingZq::from_str("4  1 0 0 1 mod 17").unwrap();
+        let poly = PolyOverZ::from_str("3  0 1 1").unwrap();
+        let mut poly_ring = PolynomialRingZq::from((&poly, &modulus));
+
+        assert!(poly_ring.set_coeff(i64::MIN, 2).is_err());
```
MoogSven, line 166: Also add a negative index check in the in the order of the degree.
```suggestion
    fn set_negative_index() {
        let modulus = ModulusPolynomialRingZq::from_str("4  1 0 0 1 mod 17").unwrap();
        let poly = PolyOverZ::from_str("3  0 1 1").unwrap();
        let mut poly_ring = PolynomialRingZq::from((&poly, &modulus));

        assert!(poly_ring.set_coeff(-1, 2).is_err());
        assert!(poly_ring.set_coeff(i64::MIN, 2).is_err());
```

MoogSven (changes requested):

src/integer/poly_over_z/set.rs
```diff
@@ -78,16 +76,16 @@ mod test_set_coeff {
         assert!(poly.set_coeff(i8::MIN, 2).is_err());
     }
 
-    /// Ensure that coefficients up to 2^15 -1 work
+    /// Ensure that big coefficients work.
     #[test]
-    fn set_max_coeff() {
+    fn set_coeff_big() {
```
MoogSven, line 81: I would prefer large over big, it is used more in other tests and is more formal. I also added this to the refactoring.

src/integer/poly_over_z/set.rs
```diff
@@ -78,16 +76,16 @@ mod test_set_coeff {
...
         let mut poly = PolyOverZ::from_str("2  1 1").unwrap();
 
         assert!(poly.set_coeff(i8::MAX, 2).is_ok());
         assert!(poly.set_coeff(i16::MAX, 2).is_ok());
     }
 
-    /// Ensure that the max of [`u8`] and [`u16`] works as a coefficient
+    /// Ensure that the max of [`u8`] and [`u16`] works as an index.
     #[test]
     fn set_unsigned_coeff() {
```
MoogSven, line 90: - [ ] Also update function name

MoogSven (approved):
I like that you also updated the other instances of set_coeff accordingly.
Please change `set_min_negative_coeff` to `set_negative_index`.

PR #374, opened 2023-08-20
Z `is_perfect_power`

jnsiemer:
**Description**
This PR implements...
- [x] is_perfect_power

for `Z`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/properties.rs
+++ b/src/integer/z/properties.rs
@@ -196,6 +196,9 @@ impl Z {
     /// Computes a pair `(root, exp)` s.t. `root^exp = self`
     /// if `self` is a perfect power and can be represented via `root.pow(exp)`.
     ///
+    /// This algorithm tries to find the smallest perfect root,
+    /// but there is no formal guarantee to find it.
+    ///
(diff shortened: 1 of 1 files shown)


src/integer/z/properties.rs
```diff
@@ -191,6 +192,176 @@ impl Z {
         unsafe { fmpz_mod(&mut out.value, &self.value, &modulus.value) };
         out
     }
+
+    /// Computes a pair `(root, exp)` s.t. `root^exp = self`
+    /// if `self` is a perfect power and can be represented via `root.pow(exp)`.
+    ///
+    /// Returns a pair `(root, exp)` if `root.pow(exp) = self` exists. Otherwise,
+    /// `None` is returned.
```
Marvin-Beckmann, line 200: What is the behaviour of 16 = 4^2 = (-4)^2= 2^4 = (-2)^4? which return can one expect, the documentation itself should make this clear.
Maybe clarify that only some pair is returned or something similar, up to your personal choice

Marvin-Beckmann (changes requested):

Marvin-Beckmann (approved):

PR #375, opened 2023-08-22
Inverse for MatZq

Marcel583:
**Description**

This PR implements the Inverse function for MatZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_zq/inverse.rs
+++ b/src/integer_mod_q/mat_zq/inverse.rs
@@ -11,11 +11,65 @@
 use super::MatZq;
 use crate::{
     integer::{MatZ, Z},
-    traits::{Concatenate, GetNumRows},
+    traits::{Concatenate, Gcd, GetNumRows},
 };
 use flint_sys::fmpz_mod_mat::fmpz_mod_mat_rref;
 
 impl MatZq {
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant co-prime to the modulus) and `None` otherwise.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer_mod_q::MatZq;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
+    ///
+    /// let matrix_invert = matrix.inverse().unwrap();
+    ///
+    /// let id = &matrix_invert * &matrix;
+    /// assert_eq!("[[5, 1],[5, 3]] mod 7", matrix_invert.to_string());
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", id.to_string());
+    /// ```
+    pub fn inverse(&self) -> Option<MatZq> {
+        // Check if the matrix is square and compute the determinant.
+        if let Ok(det) = MatZ::from(self).det() {
+            // Check weather the square matrix is invertible or not.
+            if det.gcd(self.get_mod()) != Z::ONE {
+                None
+            } else {
+                let dimensions = self.get_num_rows();
+                let mut inverse = MatZq::new(dimensions, dimensions, self.get_mod());
+
+                // The i`th unit vector in the for loop.
+                let mut e_i = MatZq::identity(dimensions, 1, self.get_mod());
+
+                // Use solve for all unit vectors except the last one.
+                for i in 0..dimensions - 1 {
+                    if let Some(column_i) = self.solve_gaussian_elimination(&e_i) {
+                        inverse.set_column(i, &column_i, 0).unwrap();
+                    } else {
+                        return None;
+                    }
+
+                    e_i.swap_entries(i, 0, i + 1, 0).unwrap();
+                }
+
+                // Use solve for the last unit vector.
+                if let Some(column_i) = self.solve_gaussian_elimination(&e_i) {
+                    inverse.set_column(dimensions - 1, &column_i, 0).unwrap();
+                } else {
+                    return None;
+                }
+
+                Some(inverse)
+            }
+        } else {
+            None
+        }
+    }
+
     /// Returns the inverse of the matrix if it exists (is square and
     /// has a determinant unequal to zero) and `None` otherwise.
     ///
@@ -43,7 +97,7 @@ impl MatZq {
             "The modulus of the matrix is not prime"
         );
 
-        // Check if the matrix is spare and compute determinant
+        // Check if the matrix is square and compute the determinant.
         if let Ok(det) = MatZ::from(self).det() {
             if det == Z::ZERO {
                 None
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -11,11 +11,65 @@
...
+
+                    e_i.swap_entries(i, 0, i + 1, 0).unwrap();
+                }
+
+                // Use solve for the last unit vector.
+                if let Some(column_i) = self.solve_gaussian_elimination(&e_i) {
+                    inverse.set_column(dimensions - 1, &column_i, 0).unwrap();
+                } else {
+                    return None;
+                }
```
Marvin-Beckmann, line 64: why did you not include it in the loop ?

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -105,11 +159,105 @@ impl MatZq {
...
+    use crate::{
+        integer::Z,
+        integer_mod_q::{MatZq, Modulus},
+    };
     use std::str::FromStr;
 
     /// Test whether `inverse` correctly calculates an inverse matrix.
     #[test]
+    fn inverse_works() {
+        let mat = MatZq::from_str("[[5,2],[2,1]] mod 14").unwrap();
```
Marvin-Beckmann, line 171: spaces between entries

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -11,11 +11,65 @@
...
+    /// let matrix_invert = matrix.inverse().unwrap();
+    ///
+    /// let id = &matrix_invert * &matrix;
+    /// assert_eq!("[[5, 1],[5, 3]] mod 7", matrix_invert.to_string());
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", id.to_string());
+    /// ```
+    pub fn inverse(&self) -> Option<MatZq> {
+        // Check if the matrix is square and compute the determinant.
+        if let Ok(det) = MatZ::from(self).det() {
+            // Check weather the square matrix is invertible or not.
```
Marvin-Beckmann, line 38:
```suggestion
            // Check whether the square matrix is invertible or not.
```

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -11,11 +11,65 @@
...
 impl MatZq {
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant co-prime to the modulus) and `None` otherwise.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer_mod_q::MatZq;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
```
Marvin-Beckmann, line 27: dont forget the spaces in the doc tests

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -105,11 +159,105 @@ impl MatZq {
...
+        integer_mod_q::{MatZq, Modulus},
+    };
     use std::str::FromStr;
 
     /// Test whether `inverse` correctly calculates an inverse matrix.
     #[test]
+    fn inverse_works() {
+        let mat = MatZq::from_str("[[5,2],[2,1]] mod 14").unwrap();
+
+        let inv = mat.inverse().unwrap();
```
Marvin-Beckmann, line 173: can you add a test where you already provdide an identity matrix?

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -11,11 +11,65 @@
...
+    /// use qfall_math::integer_mod_q::MatZq;
+    /// use std::str::FromStr;
+    ///
+    /// let mut matrix = MatZq::from_str("[[1,2],[3,4]] mod 7").unwrap();
+    ///
+    /// let matrix_invert = matrix.inverse().unwrap();
+    ///
+    /// let id = &matrix_invert * &matrix;
+    /// assert_eq!("[[5, 1],[5, 3]] mod 7", matrix_invert.to_string());
+    /// assert_eq!("[[1, 0],[0, 1]] mod 7", id.to_string());
```
Marvin-Beckmann, line 33:
```suggestion
    /// assert!(id.is_identity());
```

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -105,11 +159,105 @@ impl MatZq {
...
+    /// Check if the multiplication of inverse and matrix result in an identity matrix.
+    #[test]
+    fn inverse_correct() {
+        let mat = MatZq::from_str("[[5,2],[2,1]] mod 8").unwrap();
+
+        let inv = mat.inverse().unwrap();
+
+        let diag = mat * inv;
+        let cmp = MatZq::identity(2, 2, 8);
+        assert_eq!(cmp, diag);
```
Marvin-Beckmann, line 188: better use the is_identity function for the asserts

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -105,11 +159,105 @@ impl MatZq {
...
-    use crate::integer_mod_q::MatZq;
+    use crate::{
+        integer::Z,
+        integer_mod_q::{MatZq, Modulus},
+    };
     use std::str::FromStr;
 
     /// Test whether `inverse` correctly calculates an inverse matrix.
     #[test]
+    fn inverse_works() {
```
Marvin-Beckmann, line 170: can you add a test for a large modulus and large entries?

src/integer_mod_q/mat_zq/inverse.rs
```diff
@@ -11,11 +11,65 @@
...
 use crate::{
     integer::{MatZ, Z},
-    traits::{Concatenate, GetNumRows},
+    traits::{Concatenate, Gcd, GetNumRows},
 };
 use flint_sys::fmpz_mod_mat::fmpz_mod_mat_rref;
 
 impl MatZq {
+    /// Returns the inverse of the matrix if it exists (is square and
+    /// has a determinant co-prime to the modulus) and `None` otherwise.
```
Marvin-Beckmann, line 20: Is this a required property, I have not found it anywhere yet.
How does this behave for invertible matrices with negative determinant?, e.g. [[0, 1],[1, 0]]?

Marvin-Beckmann (changes requested):

src/integer_mod_q/mat_zq/inverse.rs
Marcel583: Because swap_entries should not be computed after the last vector, so I could either test in every loop if we are on the last vector or I could exclude the computation for the last vector like above.
Since the latter is more efficient (but has more code) I chose that implementation

src/integer_mod_q/mat_zq/inverse.rs
Marcel583: Yes, since AB = I ⟹ det(A) det(B) = 1 mod p. Negative values are treated like non negative values

src/integer_mod_q/mat_zq/inverse.rs
Marvin-Beckmann: considering the runtime of solve, and the very fast swap entries runtime, this seems like a very very low overhead, I would probably change it, even though it is slightly more ineffcient.
But do as you prefer in that case, the other just looks cleaner

Marvin-Beckmann (approved):

PR #377, opened 2023-08-22
ModPolyRingZq `new_cyclic` and `new_anticyclic`

jnsiemer:
**Description**
This PR implements...
- [x] new_cyclic
- [x] new_anticyclic

for/ of `ModulusPolynomialRingZq`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
@@ -13,9 +13,13 @@
 //! The explicit functions contain the documentation.
 
 use super::ModulusPolynomialRingZq;
-use crate::{error::MathError, integer_mod_q::PolyOverZq};
+use crate::{
+    error::MathError,
+    integer_mod_q::{Modulus, PolyOverZq},
+    traits::SetCoefficient,
+};
 use flint_sys::fq::fq_ctx_init_modulus;
-use std::{ffi::CString, mem::MaybeUninit, rc::Rc, str::FromStr};
+use std::{ffi::CString, fmt::Display, mem::MaybeUninit, rc::Rc, str::FromStr};
 
 impl From<&PolyOverZq> for ModulusPolynomialRingZq {
     /// Create a new Modulus object of type [`ModulusPolynomialRingZq`]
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -72,6 +76,70 @@ impl From<&ModulusPolynomialRingZq> for ModulusPolynomialRingZq {
...
+    /// # Panics ...
+    /// - if the `modulus` is not larger than `1`.
+    pub fn new_anticyclic(
+        n: impl TryInto<i64> + Display,
+        modulus: impl Into<Modulus>,
+    ) -> Result<Self, MathError> {
+        let mut poly = PolyOverZq::from((1, modulus));
+        poly.set_coeff(n, 1)?;
+        Ok(ModulusPolynomialRingZq::from(&poly))
+    }
```
Marvin-Beckmann, line 109: This modulus only indirectly defines an anti-cyclic lattice (i.e. using the coefficeint embedding) otherwise this only defines an anti-cyclic ideal.

src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -72,6 +76,70 @@ impl From<&ModulusPolynomialRingZq> for ModulusPolynomialRingZq {
...
+    pub fn new_cyclic(
+        n: impl TryInto<i64> + Display,
+        modulus: impl Into<Modulus>,
+    ) -> Result<Self, MathError> {
+        let mut poly = PolyOverZq::from((-1, modulus));
+        poly.set_coeff(n, 1)?;
+        Ok(ModulusPolynomialRingZq::from(&poly))
+    }
+}
+
```
Marvin-Beckmann, line 142: see above

src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -99,6 +167,126 @@ impl FromStr for ModulusPolynomialRingZq {
     }
 }
 
+#[cfg(test)]
```
Marvin-Beckmann, line 170: please also compare the entries of the polynomial, such that the first coefficient is 1/-1 and the last is 1

Marvin-Beckmann (changes requested):
The functionality on its own is a nice to have, but I do not think, that it should be in the math crate.
e.g. I also included the rotation matrices in the crypto crate.
I personally prefer them to be in the crypto crate, specifically if we are explicitly talking about default lattices here.

jnsiemer:
Contents were moved to `utils::common_moduli` module in crypto-crate. See https://github.com/qfall/crypto/pull/43

PR #378, opened 2023-08-22
Different parameter types for `Z::div_(floor, ceil, exact)`

MoogSven:
**Description**
just small addition quality of live improvement for another PR. 

This PR implements...
- Into<Z> parameters for div_floor, div_ceil, and div_exact

for/ of `Component`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/arithmetic/div.rs
+++ b/src/integer/z/arithmetic/div.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Niklas Siemer
+// Copyright © 2023 Niklas Siemer and Sven Moog
 //
 // This file is part of qFALL-math.
 //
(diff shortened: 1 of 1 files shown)


Marvin-Beckmann (commented):
I left out changing these in the refactoeing, as the standard implementation as you propose would mean, that we always have to create an owned version, i. e. clone, do you see a way withou having to do that?

MoogSven:
> I left out changing these in the refactoeing, as the standard implementation as you propose would mean, that we always have to create an owned version, i. e. clone, do you see a way without having to do that?

We could use the AsInteger trait to get a reference when a reference is easily available, e.g. &Z parameter. That would avoid the clone. But that is more work. I don't think that these functions are called that often to be worth spending the time to optimize them at this stage of the project.

These changes would probably only improve the performance when passing &Z or &Modulus. (But I don't know how much)

Marvin-Beckmann (approved):

PR #380, opened 2023-08-22
Z `sample_binomial`

jnsiemer:
**Description**
This PR implements...
- [x] sample_binomial

for `Z`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/sample/binomial.rs
+++ b/src/integer/z/sample/binomial.rs
@@ -21,7 +21,7 @@ impl Z {
     ///
     /// Returns a fresh [`Z`] instance with a value sampled
     /// according to the binomial distribution or a [`MathError`]
-    /// if `n < 1`, `p <= 0`, `p >= 1`, or `n` does not fit into an [`i64`].
+    /// if `n < 1`, `p ∉ (0,1)`, or `n` does not fit into an [`i64`].
     ///
     /// # Examples
     /// ```
@@ -34,7 +34,7 @@ impl Z {
     /// - Returns a [`MathError`] of type [`InvalidIntegerInput`](MathError::InvalidIntegerInput)
     /// if `n < 1`.
     /// - Returns a [`MathError`] of type [`InvalidInterval`](MathError::InvalidInterval)
-    /// if `p <= 0` or `p >= 1`.
+    /// if `p ∉ (0,1)`.
     /// - Returns a [`MathError`] of type [`ConversionError`](MathError::ConversionError)
     /// if `n` does not fit into an [`i64`].
     pub fn sample_binomial(n: impl Into<Z>, p: impl Into<Q>) -> Result<Self, MathError> {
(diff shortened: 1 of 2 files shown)


src/integer/z/sample/binomial.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Chooses a [`Z`] instance according to the binomial distribution
+    /// parameterized by `n` and `p`.
+    ///
+    /// Parameters:
+    /// - `n`: specifies the number of trials
+    /// - `p`: specifies the probability of success
+    ///
+    /// Returns a fresh [`Z`] instance with a value sampled
+    /// according to the binomial distribution or a [`MathError`]
+    /// if `n < 1`, `p <= 0`, `p >= 1`, or `n` does not fit into an [`i64`].
```
Marvin-Beckmann, line 24:
```suggestion
    /// if `n < 1`, `p ∉ (0,1)`, or `n` does not fit into an [`i64`].
```

src/integer/z/sample/binomial.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// use qfall_math::integer::Z;
+    ///
+    /// let sample = Z::sample_binomial(2, 0.5).unwrap();
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`InvalidIntegerInput`](MathError::InvalidIntegerInput)
+    /// if `n < 1`.
+    /// - Returns a [`MathError`] of type [`InvalidInterval`](MathError::InvalidInterval)
+    /// if `p <= 0` or `p >= 1`.
```
Marvin-Beckmann, line 37:
```suggestion
    /// if `p ∉ (0,1)`.
```

src/integer/z/sample/binomial.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// - Returns a [`MathError`] of type [`ConversionError`](MathError::ConversionError)
+    /// if `n` does not fit into an [`i64`].
+    pub fn sample_binomial(n: impl Into<Z>, p: impl Into<Q>) -> Result<Self, MathError> {
+        let n: Z = n.into();
+        let p: Q = p.into();
+
+        let sample = sample_binomial(&n, &p)?;
+        Ok(Z::from(sample))
```
Marvin-Beckmann, line 45: Please consider the following change, just now came to my mind: I would let the binomial sampler (the intern function return a `Z` value rather than a `u64` -> This makes changing it later much easier as only the underlying function has to change and not each function calling it

Marvin-Beckmann (approved):

src/integer/z/sample/binomial.rs
jnsiemer: That's a change, I'm going to make once we want to sample values larger than u64::MAX. Currently, it is just more efficient to return the u64 value and transform it into the needed value.

PR #404, opened 2024-03-17
Cholesky decomposition

Marvin-Beckmann:
**Description**

This PR implements the Cholesky decomposition as a placeholder until we can use the function directly in FLINT through the FFI (not yet supported by FFI).
This implementation gives a usable interface, provides respective tests and allows for further implementations that require this algorithm.

**Testing**

**Checklist:**


src/rational/mat_q/cholesky_decomp.rs
```diff
@@ -0,0 +1,151 @@
...
+            let mut l_i = MatQ::identity(n, n);
+            l_i.set_column(i, &column_a_i, 0).unwrap();
+            l = l * l_i;
+
+            // update matrix A recursively
+            if i < n - 1 {
+                let b = a.get_submatrix(1, n - i - 1, 1, n - i - 1).unwrap();
+                let b_minus = (1 / a_ii)
+                    * a.get_submatrix(1, n - i - 1, 0, 0).unwrap()
+                    * a.get_submatrix(0, 0, 1, n - i - 1).unwrap();
```
LPorzenheim, line 79: Matrix is symmetric, is transpose faster than get_submatrix?

src/rational/mat_q/cholesky_decomp.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Ensure that the function panics if a non-square matrix is provided
+    #[test]
+    #[should_panic]
+    fn non_square() {
+        let matrix = MatQ::new(3, 2);
+
+        matrix.cholesky_decomposition();
+    }
+
+    /// Ensure that the function panics if a non-positive-definite matrix is provided
```
LPorzenheim, line 113: Does not match documentation, since that states that function does not check for pos. def. Function only checks for diag>0

LPorzenheim (changes requested):

src/rational/mat_q/cholesky_decomp.rs
Marvin-Beckmann: yes, somewhow overlooked that our sqrt panicks if provided with negative values. Fixed the documentation that the function captures this

src/rational/mat_q/cholesky_decomp.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// assert_eq!(matrix, &l * l.transpose());
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if `self` is not a symmetric matrix.
+    ///
+    /// # Note
+    /// - the function may behave unintended if the provided matrix is not a
+    ///     Hermitian positive-definite matrix.
+    pub fn cholesky_decomposition(&self) -> MatQ {
```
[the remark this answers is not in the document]
Marvin-Beckmann: added it in description (stating that its the cholesky algorithm)

src/rational/mat_q/cholesky_decomp.rs
Marvin-Beckmann: if one of them is faster, then it should be get_submatrix - I think.

Both utilize the underlying Flint-functions: transpose and window respectively.

transpose iterates through all entries and sets the entries respectively - therefore we have the additional iteration.
the window first gets pointers for each entry and then simply performs a deepclone of all entries in the range.

In the end -> both should be basically identical

LPorzenheim (approved):
Looks good, maybe add a link to the Cholesky algorithm

PR #407, opened 2024-09-02
SampleD with a non-spherical Gaussian convolution matrix (with standard base and center 0)

Marvin-Beckmann:
**Description**

This PR implements...
- [x] SampleD with a non-spherical Gaussian convolution matrix (with standard base and center 0)

I decided to only implement the common version for now, as the more general version requires more extensive testing and as of right now, I only need the common version.

The PR is based on Algorithm 1 from <https://link.springer.com/chapter/10.1007/978-3-642-14623-7_5>.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_z/sample/discrete_gauss.rs
+++ b/src/integer/mat_z/sample/discrete_gauss.rs
@@ -12,7 +12,7 @@ use crate::{
     error::MathError,
     integer::{MatZ, Z},
     rational::{MatQ, Q},
-    traits::{GetNumColumns, GetNumRows, Pow, SetEntry},
+    traits::{GetNumColumns, GetNumRows, SetEntry},
     utils::sample::discrete_gauss::{sample_d, sample_d_precomputed_gso, sample_z},
 };
 use std::fmt::Display;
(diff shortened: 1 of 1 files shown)


src/integer/mat_z/sample/discrete_gauss.rs
```diff
@@ -306,3 +368,88 @@ mod test_sample_d {
...
+    #[test]
+    #[should_panic]
+    fn no_convolution_matrix_1() {
+        let r = Q::from(4);
+        let convolution_matrix = MatQ::from_str("[[-1,1],[1,1]]").unwrap();
+        let convolution_matrix = (convolution_matrix - r.pow(2).unwrap() * MatQ::identity(2, 2))
+            .cholesky_decomposition();
+
+        let _ = MatZ::sample_d_common_non_spherical(16, &convolution_matrix, r).unwrap();
+    }
```
jnsiemer, line 418: Please, make this an error too and return it, if it happens. Furthermore, it needs to be added to the doc-comment.

jnsiemer (changes requested):
Please consider the proposed changes - especially regarding panics and errors.
Otherwise, great work! I won't trace every call back but this algorithm should work as expected.

jnsiemer (approved):

PR #408, opened 2024-09-05
Refactor Froms

Marcel583:
**Description**

This PR deletes unused functions and fixes the documentation in from files.

**Checklist:**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,7 +83,7 @@ jobs:
         run: cargo outdated --root-deps-only --exit-code 1
 
       - name: Archive release build
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
         with:
           name: ${{ env.PROJECT_NAME }}-release_build-v${{ env.PROJECT_VERSION }}
           path: target/release/${{ env.PROJECT_NAME }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@ name: Pipeline
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
 
 env:
   CARGO_TERM_COLOR: always
@@ -12,11 +12,11 @@ jobs:
   full_pipeline:
     name: Full Pipeline
     runs-on: ubuntu-latest
-    
+
     steps:
       - uses: actions/checkout@v3
 
-      - name: Setup dtolnay/rust-toolchain 
+      - name: Setup dtolnay/rust-toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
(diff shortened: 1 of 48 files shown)


.github/workflows/main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Calculate test coverage
         run: cargo tarpaulin --out Html
       - name: Archive code coverage results
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
```
Marvin-Beckmann, line 60: is there a specific reason for changing the version?

src/integer/mat_z/from.rs
```diff
@@ -80,9 +77,10 @@ impl FromStr for MatZ {
     /// # Errors and Failures
     /// - Returns a [`MathError`] of type [`StringConversionError`](MathError::StringConversionError)
     ///     - if the matrix is not formatted in a suitable way,
-    ///     - if the number of entries in rows is unequal,
-    ///     - if an entry contains a Nul byte, or
+    ///     - if the number of rows or columns is too large (must fit into i64),
+    ///     - if the number of entries in rows is unequal, or
     ///     - if an entry is not formatted correctly.
+    ///     - For further information see [`Z::from_str`].
```
Marvin-Beckmann, line 83: this formatting only relates to Z::from_str -> hence the statement "For further information..." should be on that dedicated bullet point. Otherwise it might be confusing

src/integer_mod_q/mat_zq/from.rs
```diff
@@ -130,9 +100,31 @@ impl FromStr for MatZq {
...
+    /// let m = MatZ::from_str("[[1, 2],[3, -1]]").unwrap();
+    ///
+    /// let a = MatZq::from((&m, 17));
+    /// ```
     fn from((matrix, modulus): (&MatZ, Mod)) -> Self {
-        MatZq::from_mat_z_modulus(matrix, modulus)
+        let mut out = MatZq::new(matrix.get_num_rows(), matrix.get_num_columns(), modulus);
+        unsafe {
+            fmpz_mat_set(&mut out.matrix.mat[0], &matrix.matrix);
+            _fmpz_mod_mat_reduce(&mut out.matrix);
```
Marvin-Beckmann, line 125: Consider changing it to `fmpz_mod_mat_set_fmpz_mat` (should do the same and we have less interface usage)

Marvin-Beckmann (changes requested):

.github/workflows/main.yml
Marcel583: The pipeline fails now because version 1 is outdated. Switching to version 3 fixes the problem.

Marvin-Beckmann (approved):

PR #409, opened 2024-09-05
Get_mat for MatZq

Marcel583:
**Description**

This PR implements get_mat for MatZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_zq/get.rs
+++ b/src/integer_mod_q/mat_zq/get.rs
@@ -27,11 +27,12 @@ use flint_sys::{
 use std::{fmt::Display, mem::MaybeUninit};
 
 impl MatZq {
-    /// Create a [`MatZ`] from a [`MatZq`].
+    /// Creates a [`MatZ`] where each entry is the representative of the 
+    /// equivalence class of each entry from a [`MatZq`].
     ///
-    /// The output matrix entries values will be in the range `[0, Modulus)`.
-    /// Use [`MatZq::get_closest_to_zero_representative`] if they should be
-    /// close to `0`.
+    /// The values in the output matrix are in the range of `[0, Modulus)`.
+    /// Use [`MatZq::get_closest_to_zero_representative`] if they should be 
+    /// in the range `[-Modulus/2, Modulus/2]`.
     ///
     /// Returns the matrix as a [`MatZ`].
     ///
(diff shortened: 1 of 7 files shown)


src/integer_mod_q/mat_zq/get.rs
```diff
@@ -26,139 +27,43 @@ use flint_sys::{
 use std::{fmt::Display, mem::MaybeUninit};
 
 impl MatZq {
-    /// Returns the modulus of the matrix as a [`Modulus`].
+    /// Create a [`MatZ`] from a [`MatZq`].
```
Marvin-Beckmann, line 30: Create might be a bit confusing. Maybe more: Creates a MatZ where each entry is the representative of the equivalence class of each entry.

or similar

src/integer_mod_q/mat_zq/get.rs
```diff
@@ -26,139 +27,43 @@ use flint_sys::{
...
     ///
-    /// # Examples
-    /// ```
-    /// use qfall_math::integer_mod_q::MatZq;
+    /// The output matrix entries values will be in the range `[0, Modulus)`.
+    /// Use [`MatZq::get_closest_to_zero_representative`] if they should be
+    /// close to `0`.
```
Marvin-Beckmann, line 34: "if they should be close to `0`" is unclear what you mean...

src/integer_mod_q/mat_zq/get.rs
```diff
@@ -26,139 +27,43 @@ use flint_sys::{
...
-    /// use qfall_math::integer::Z;
     /// use std::str::FromStr;
     ///
-    /// let matrix = MatZq::from_str("[[1, 2, 3],[4, 5, 6],[7, 8, 9]] mod 10").unwrap();
+    /// let m = MatZq::from_str("[[1, 2],[3, -1]] mod 5").unwrap();
     ///
-    /// assert_eq!(Z::from(3), matrix.get_entry(0, 2).unwrap());
-    /// assert_eq!(Z::from(8), matrix.get_entry(2, 1).unwrap());
-    /// assert_eq!(Z::from(8), matrix.get_entry(-1, -2).unwrap());
+    /// let a = m.get_mat();
```
Marvin-Beckmann, line 46: maybe add an assertion with the expected matrix so the person looking at the example can verify that the -1 will be 'changed' to a 4

Marvin-Beckmann (approved):
Thanks for the implementation of the change. Just a small change in the documentation is needed

PR #410, opened 2024-09-10
Add z poly z

3pmile:
**Description**

Addition for Z and PolyOverZ

This PR implements...
- [x]  Z + PolyOverZ (and all burrowed versions)
- [ ]  PolyOverZ + Z (and all burrowed versions)

for/ of `Component`.

Addition

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/arithmetic/add.rs
+++ b/src/integer/z/arithmetic/add.rs
@@ -490,6 +490,19 @@ mod test_add_between_z_and_poly_over_z {
         assert_eq!(c, PolyOverZ::from_str("4  10 1 2 3").unwrap());
     }
 
+    /// Testing addition for large numbers
+    #[test]
+    fn add_large_numbers() {
+        let a: Z = Z::from(i64::MAX);
+        let b: PolyOverZ =
+            PolyOverZ::from_str(&format!("3  {} {} {}", i64::MAX, u64::MAX, i32::MAX)).unwrap();
+        let c: PolyOverZ = a + b;
+        assert_eq!(
+            c,
+            PolyOverZ::from_str(&format!("3  {} {} {}", u64::MAX - 1, u64::MAX, i32::MAX)).unwrap()
+        );
+    }
+
     /// Testing addition for an empty polynomial
     #[test]
     fn add_zero_polynomial() {
--- a/src/integer/z/arithmetic/add.rs
+++ b/src/integer/z/arithmetic/add.rs
@@ -10,6 +10,7 @@
 
 use super::super::Z;
 use crate::{
+    integer::PolyOverZ,
     integer_mod_q::Zq,
     macros::arithmetics::{
         arithmetic_between_types, arithmetic_trait_borrowed_to_owned,
@@ -21,6 +22,9 @@ use flint_sys::{
     fmpq::fmpq_add_fmpz,
     fmpz::{fmpz, fmpz_add},
     fmpz_mod::fmpz_mod_add_fmpz,
+    fmpz_poly::{
+        fmpz_poly_get_coeff_fmpz, fmpz_poly_set, fmpz_poly_set_coeff_fmpz, fmpz_poly_struct,
+    },
 };
 use std::ops::Add;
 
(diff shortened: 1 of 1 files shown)


3pmile:
Please have a look at the implementation and whether edge cases for tests are missing

src/integer/z/arithmetic/add.rs
```diff
@@ -140,6 +144,51 @@ impl Add<&Zq> for &Z {
...
+            length: 0,
+        };
+        unsafe {
+            fmpz_poly_set(&mut out, &other.poly);
+            fmpz_poly_get_coeff_fmpz(&mut added, &other.poly, 0);
+            fmpz_add(&mut added, &self.value, &added);
+            fmpz_poly_set_coeff_fmpz(&mut out, 0, &added);
+        }
+        PolyOverZ { poly: out }
+    }
```
Marvin-Beckmann, line 186:
```suggestion
        // check if the first coefficient of the polynomial is initiated and
        // can be addressed
        if other.is_zero() {
            PolyOverZ::from(self)
        } else {
            let out = other.clone();
            unsafe {
                fmpz_add(out.poly.coeffs, &self.value, out.poly.coeffs);
            }
            out
        }
```

I think this solution is cleaner as we only have to interact with FLINT itslef minimally. Also, your `added` is never cleared which would cause a memory-leak

Marvin-Beckmann: poly.coeffs adresses the first fmpz in the list of coefficients, therefore we can address the first coefficient this way (if defined)

src/integer/z/arithmetic/add.rs
```diff
@@ -21,6 +22,9 @@ use flint_sys::{
     fmpq::fmpq_add_fmpz,
     fmpz::{fmpz, fmpz_add},
     fmpz_mod::fmpz_mod_add_fmpz,
+    fmpz_poly::{
+        fmpz_poly_get_coeff_fmpz, fmpz_poly_set, fmpz_poly_set_coeff_fmpz, fmpz_poly_struct,
+    },
```
Marvin-Beckmann, line 27: probably dont need all when adapting the changes

src/integer/z/arithmetic/add.rs
```diff
@@ -398,3 +447,68 @@ mod test_add_between_z_and_zq {
...
+    }
+
+    /// Testing addition for [`Z`] and borrowed [`PolyOverZ`]
+    #[test]
+    fn add_second_borrowed() {
+        let a: Z = Z::from(9);
+        let b: PolyOverZ = PolyOverZ::from_str("4  1 1 2 3").unwrap();
+        let c: PolyOverZ = a + &b;
+        assert_eq!(c, PolyOverZ::from_str("4  10 1 2 3").unwrap());
+    }
```
Marvin-Beckmann, line 491: possibly batch them for brevity and call the test itself availability as we have done for some other types

@Marcel583 can probably look for what is most consistent, but I would opt for a less repetititve design (at least right now I would)

src/integer/z/arithmetic/add.rs
```diff
@@ -398,3 +447,68 @@ mod test_add_between_z_and_zq {
...
+
+    /// Testing addition for an empty polynomial
+    #[test]
+    fn add_zero_polynomial() {
+        let a: Z = Z::from(15);
+        let b: PolyOverZ = PolyOverZ::default();
+        let c: PolyOverZ = a + b;
+        assert_eq!(c, PolyOverZ::from_str("1  15").unwrap());
+    }
```
Marvin-Beckmann, line 513: I would also add the edgecase where you add a Z::ZERO (for consistency in both directions)

src/integer/z/arithmetic/add.rs
```diff
@@ -140,6 +144,51 @@ impl Add<&Zq> for &Z {
...
+    /// use std::str::FromStr;
+    ///
+    /// let a: Z = Z::from(42);
+    /// let b: PolyOverZ = PolyOverZ::from_str("4  5 1 2 3").unwrap();
+    ///
+    /// let c: PolyOverZ = &a + &b;
+    /// let d: PolyOverZ = a + b;
+    /// let e: PolyOverZ = &Z::from(42) + d;
+    /// let f: PolyOverZ = Z::from(42) + &e;
+    /// ```
```
Marvin-Beckmann, line 170: maybe add an `assert_eq!(...)` where the reader can see the actual addition
```suggestion
    /// ```
    /// use qfall_math::integer::PolyOverZ;
    /// use qfall_math::integer::Z;
    /// use std::str::FromStr;
    ///
    /// let a: Z = Z::from(42);
    /// let b: PolyOverZ = PolyOverZ::from_str("4  5 1 2 3").unwrap();
    ///
    /// let c: PolyOverZ = &a + &b;
    /// let d: PolyOverZ = a + b;
    /// let e: PolyOverZ = &Z::from(42) + d;
    /// let f: PolyOverZ = Z::from(42) + &e;
    /// 
    /// assert_eq!(PolyOverZ::from_str!("4  47 1 2 3").unwrap(), a+b)
    /// ```
```
Maybe choose some clever way so the expected behaviour is shown and not just the available types

src/integer/z/arithmetic/add.rs
3pmile: applied

src/integer/z/arithmetic/add.rs
3pmile: cleared

src/integer/z/arithmetic/add.rs
3pmile: added

src/integer/z/arithmetic/add.rs
3pmile: I have talked to @Marcel583 who said this is the consistent solution for now, but we should have a look at all test in a refactoring

src/integer/z/arithmetic/add.rs
3pmile: in alinement with the other tests I will keep the current implementation

Marvin-Beckmann (approved):

PR #412, opened 2024-09-20
Make `Q::simplify` more viable

jnsiemer:
**Description**

The function `simplify` is capable of more than we thought. This PR makes sure that any developer is aware of the possibility to use it as a trade-off between precision and memory usage.

This PR implements...
- [x] new constants for `Q` : `Q::MAX62` (as this is the maximum number that we can instantiate as a constant and fmpz), `INV_MAX62`, `INV_MAX32` as constants to use for a fair trade-off reg precision and memory-usage.
- [x] explanations and an example of how to use `simplify` with these constants
- [x] ensurance that the sign does not change using `simplify`

for `Q`.

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/q/default.rs
+++ b/src/rational/q/default.rs
@@ -111,6 +111,51 @@ impl Q {
             den: fmpz(842468587426513207),
         },
     };
+
+    /// Returns an instantiation of [`Q`] with value `2^62 - 1`.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::rational::Q;
+    ///  
+    /// let a: Q = Q::MAX62;
+    /// ```
+    pub const MAX62: Q = Q {
+        value: fmpq {
+            num: fmpz(i64::pow(2, 62) - 1),
+            den: fmpz(1),
+        },
+    };
+
+    /// Returns an instantiation of [`Q`] with value `1 / (2^62 - 1)`.
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::rational::Q;
+    ///  
+    /// let a: Q = Q::INV_MAX62;
+    /// ```
+    pub const INV_MAX62: Q = Q {
+        value: fmpq {
+            num: fmpz(1),
+            den: fmpz(i64::pow(2, 62) - 1),
+        },
+    };
+
+    /// Returns an instantiation of [`Q`] with value `1 / (2^32 - 1)`.
+    ///
+    /// # Examples
+    /// ```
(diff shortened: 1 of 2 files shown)


Marvin-Beckmann (changes requested):
I like the changes, but two more suggestions:
- also add a constant for 16 and maybe even 8 bits
- add a test for the simplify that ensures that the denominator is actually at most having that specific number of bits when we have provided it.ideally also for large rationals and such

src/rational/q/rounding.rs
```diff
@@ -264,14 +291,38 @@ mod test_simplify {
...
     }
 
+    /// Ensure max_bits of denominator are not bigger than 1/2 * precision
+    #[test]
+    fn max_bits_denominator() {
+        let value = Q::PI;
+        let precisions = [Q::INV_MAX8, Q::INV_MAX16, Q::INV_MAX32];
+
+        for precision in precisions {
+            let inv_precision = precision.inverse().unwrap().get_numerator();
```
Marvin-Beckmann, line 301:
```suggestion
            let inv_precision = precision.get_denominator();
```

src/rational/q/rounding.rs
```diff
@@ -109,15 +116,35 @@ impl Q {
...
-        let upper = self + precision;
+    pub fn simplify(&self, precision: impl Into<Q>) -> Self {
+        let precision = precision.into();
+
+        let lower = self - &precision;
+        let upper = self + &precision;
         let mut out = Q::default();
         unsafe { fmpq_simplest_between(&mut out.value, &lower.value, &upper.value) };
+
+        if unsafe { fmpq_sgn(&self.value) != fmpq_sgn(&out.value) } {
```
Marvin-Beckmann, line 144: we could consider adding the sign function to our types

src/rational/q/default.rs
```diff
@@ -111,6 +111,81 @@ impl Q {
...
+    /// use qfall_math::rational::Q;
+    ///  
+    /// let a: Q = Q::INV_MAX8;
+    /// ```
+    pub const INV_MAX8: Q = Q {
+        value: fmpq {
+            num: fmpz(1),
+            den: fmpz(i64::pow(2, 8) - 1),
+        },
+    };
```
Marvin-Beckmann, line 188: If we add the inverse, I would also add the non-inverse - otherwise this might seem strange to users.

Marvin-Beckmann (approved):

PR #415, opened 2024-09-25
Cholesky Decomposition with less Precision

Marvin-Beckmann:
**Description**

Problem: The current implementation of the cholesky decomposition is far to slow as
- We use a self-implementation
- We are working with [`Q`]s which have a high precision at the cost of runtime

Therefore I looked at ways to overcome these issues
- Using FLINTs function -> sadly not supported in FFI
- Use simplify to get to a certain precision -> too slow...
- [x] Use [`f64`] (double-precision) between conversions
The latest of these options proved to be the best. As we can not access FLINTs function, I basically reformulated their implementation into Rust and used the same format.

for/ of `Component`.

**Testing**

-  Due to the precision - it does not work for large integers

**Checklist:**

### DIFF
--- a/src/rational/mat_q/cholesky_decomp.rs
+++ b/src/rational/mat_q/cholesky_decomp.rs
@@ -12,13 +12,13 @@
 use super::MatQ;
 use crate::{
     rational::Q,
-    traits::{Concatenate, GetEntry, GetNumColumns},
+    traits::{Concatenate, GetEntry, GetNumColumns, GetNumRows, SetEntry},
 };
 
 impl MatQ {
     /// This function performs the Cholesky decomposition (the Cholesky algorithm) and
     /// produces a matrix `L` such that `self = L * L^T`.
-    /// This function relies on the precision of `Q::sqrt` and will not provide
+    /// This function relies on the precision of [`Q::sqrt`] and will not provide
     /// perfect results in all cases.
     /// Furthermore, the Cholesky decomposition requires a Hermitian positive-definite
     /// matrix.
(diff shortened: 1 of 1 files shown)


src/rational/mat_q/cholesky_decomp.rs
```diff
@@ -79,6 +77,65 @@ impl MatQ {
...
+    /// - if `self` is not a symmetric matrix,
+    /// - if `self` has eigenvalues smaller than `0`.
+    pub fn cholesky_decomposition_flint(&self) -> MatQ {
+        assert!(self.is_symmetric(), "The provided matrix is not symmetric.");
+
+        let mut out = MatQ::new(self.get_num_rows(), self.get_num_columns());
+
+        // This code snippet originates from [`fmpz_mat_chol_d`] (FLINT)
+        for i in 0..self.get_num_rows() {
+            for j in 0..(i + 1) {
```
jnsiemer, line 117:
```suggestion
            for j in 0..=i {
```

src/rational/mat_q/cholesky_decomp.rs
```diff
@@ -79,6 +77,65 @@ impl MatQ {
...
+                let mut s: f64 = 0.0;
+                for k in 0..j {
+                    let r_ik = f64::from(&out.get_entry(i, k).unwrap());
+                    let r_jk = f64::from(&out.get_entry(j, k).unwrap());
+                    s += r_ik * r_jk
+                }
+                if i == j {
+                    let a_ii = f64::from(&self.get_entry(i, i).unwrap());
+                    assert!(a_ii > s, "The provided matrix is not positive definite.");
```
jnsiemer, line 126: I'm not sure where this condition comes from. I couldn't confirm this in the time, I spend for this review. Can you give an explanation and attach it at this point in the code?

jnsiemer (approved):

src/rational/mat_q/cholesky_decomp.rs
Marvin-Beckmann: See e.g. here https://en.wikipedia.org/wiki/Cholesky_decomposition in the section of the cholesky algorithm - there you can see the requirement

PR #419, opened 2024-09-28
Distance for modulus and support for Zq

Marcel583:
**Description**

This PR implements the distance function for Modulus.
It also deletes the distance function for Zq and adds an alternative, which is the function "get_close_to_zero_representative", which we can use in our crypto crate alternatively. This function is analogously to our function for MatZq with the same name.

**Testing**

**Checklist:**


jnsiemer:
There is a good reason, why I have not / we have decided not to implement `distance` for any mixes of Z and Zq values.
The distance for Z values is clearly defined: |x-y|.
The distance for Zq values is clearly defined: min{|x-y mod q|, |y-x mod q|}.
But for the mix, there no clear definition. Therefore, we decided not to provide the function and I would leave it at that.
The same question arises for any mix of Modulus, Z, and Zq values. How is it defined properly?
Nevertheless, I would question whether this function is ever going to be used / useful.

src/integer_mod_q/modulus/distance.rs
```diff
@@ -0,0 +1,80 @@
...
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains the implementation of the [`Distance`] trait for [`Modulus`].
+
+use super::Modulus;
+use crate::{integer::Z, traits::Distance};
+
+impl<Integer: Into<Z>> Distance<Integer> for Modulus {
```
jnsiemer, line 14: I don't think, this is well defined. Please also remove this part.

jnsiemer (changes requested):
Went over it quickly. Please apply changes according to our last joint decision in the meeting.
There should be way fewer changes once they are applied s.t. it's easier for me to review then. Thanks :)

src/integer_mod_q/z_q/get.rs
```diff
@@ -16,23 +16,50 @@ use crate::{
...
     /// assert_eq!(Z::from(4), z_value);
     /// ```
-    pub fn get_value(&self) -> Z {
+    pub fn get_representative_0_modulus(&self) -> Z {
         self.value.clone()
     }
 
+    /// Returns the [`Z`] value of the [`Zq`] element with the representatives close to `0`.
+    ///
+    /// The output value is in the range of `[-modulus/2, modulus/2]`.
```
jnsiemer, line 39: How is this interval defined if `modulus` is odd? Which value will be rounded up and which rounded down? Please make this explicite here.

jnsiemer (approved):
After long discussion, we finally made it! Thanks for holding up so well with this one :)
There is just one minor doc comment that should be specified. Otherwise, I'm interested in how this works out in the crypto crate

PR #422, opened 2024-10-03
From for MatPolyRingZq with support for MatZ

Marcel583:
**Description**

We currently only have From for MatPolyRingZq with input MatPolyRingZq and  Into<ModulusPolynomialRingZq>. 
Since MatZ can be transformed into MatPolyRingZq, I changed our current implementation such that the input can be Into<MatPolyRingZq> and  Into<ModulusPolynomialRingZq>, without losing the efficiency of our current implementation.

This PR implements...
- [x] From for MatPolyRingZq with support for MatZ

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_polynomial_ring_zq/from.rs
+++ b/src/integer_mod_q/mat_polynomial_ring_zq/from.rs
@@ -13,15 +13,19 @@
 use super::MatPolynomialRingZq;
 use crate::{integer::MatPolyOverZ, integer_mod_q::ModulusPolynomialRingZq};
 
-impl<Mod: Into<ModulusPolynomialRingZq>> From<(&MatPolyOverZ, Mod)> for MatPolynomialRingZq {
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/mat_polynomial_ring_zq/from.rs
```diff
@@ -32,66 +36,48 @@ impl<Mod: Into<ModulusPolynomialRingZq>> From<(&MatPolyOverZ, Mod)> for MatPolyn
...
     }
 }
 
-impl From<&MatPolynomialRingZq> for MatPolynomialRingZq {
-    /// Alias for [`MatPolynomialRingZq::clone`].
-    fn from(value: &MatPolynomialRingZq) -> Self {
-        value.clone()
-    }
-}
-
```
Marvin-Beckmann, line 84: why do you remove this?

Marcel583: This was a mistake, I did not want to remove it. I added it again in the next commit

Marvin-Beckmann (approved):

PR #423, opened 2024-10-04
GetCoefficient for ModulusPolyRingZq

Marcel583:
**Description**

Since we currently can not get a coefficient from a ModulusPolyRingZq:

This PR implements...
- [ ] GetCoefficient for ModulusPolyRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
@@ -171,7 +171,11 @@ impl ModulusPolynomialRingZq {
 
 #[cfg(test)]
 mod test_get_coeff_z {
-    use crate::{integer::Z, integer_mod_q::{ModulusPolynomialRingZq, Zq}, traits::GetCoefficient};
+    use crate::{
+        integer::Z,
+        integer_mod_q::{ModulusPolynomialRingZq, Zq},
+        traits::GetCoefficient,
+    };
     use std::str::FromStr;
 
     /// Ensure that `0` is returned if the provided index is not yet set.
@@ -206,7 +210,10 @@ mod test_get_coeff_z {
                 .unwrap();
 
         assert_eq!(Z::from(u64::MAX - 1), poly.get_coeff(1).unwrap());
-        assert_eq!(Zq::from((u64::MAX - 1, u64::MAX)), poly.get_coeff(1).unwrap());
+        assert_eq!(
+            Zq::from((u64::MAX - 1, u64::MAX)),
+            poly.get_coeff(1).unwrap()
+        );
     }
 
     /// Tests if negative index yields an error in get_coeff with [`Z`].
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Marvin Beckmann, Niklas Siemer
+// Copyright © 2023 Marcel Luca Schmidt, Marvin Beckmann, Niklas Siemer
 //
 // This file is part of qFALL-math.
 //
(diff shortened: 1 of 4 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -8,11 +8,120 @@
...
+    /// let coeff_4: Zq = poly.get_coeff(4).unwrap();
+    ///
+    /// assert_eq!(Zq::from((0, 17)), coeff_0);
+    /// assert_eq!(Zq::from((1, 17)), coeff_1);
+    /// assert_eq!(Zq::from((0, 17)), coeff_4);
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`OutOfBounds`](MathError::OutOfBounds) if
+    ///     either the index is negative or it does not fit into an [`i64`].
```
Marvin-Beckmann, line 55: what happens if I try to access a coefficient that is not set?, i.e. I have a polynomial of degree 5 and I try to access position 10

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -8,11 +8,120 @@
...
+                &mut out_z.value,
+                &self.modulus.modulus[0],
+                index,
+                &self.get_fq_ctx_struct().ctxp[0],
+            )
+        }
+
+        let modulus = Modulus {
+            modulus: Rc::new(self.get_fq_ctx_struct().ctxp[0]),
+        };
```
Marvin-Beckmann, line 71: I think you introduce a memory leak here:

In modulus you have saved the value with a reference counter, but still point towards the value saved in the polynomial. Meaning: as long as the polynomial is around, the modulus will only be dropped if all instantiations of the modulus are destroyed. But if this happens, the polynomial is still there, but the referenced modulus has been dropped. The other way around also works the same way

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -8,11 +8,120 @@
...
+    /// let coeff_4: Z = poly.get_coeff(4).unwrap();
+    ///
+    /// assert_eq!(Z::ZERO, coeff_0);
+    /// assert_eq!(Z::ONE, coeff_1);
+    /// assert_eq!(Z::ZERO, coeff_4);
+    /// ```
+    ///
+    /// # Errors and Failures
+    /// - Returns a [`MathError`] of type [`OutOfBounds`](MathError::OutOfBounds) if
+    ///     either the index is negative or it does not fit into an [`i64`].
```
Marvin-Beckmann, line 108: see before

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -60,6 +169,72 @@ impl ModulusPolynomialRingZq {
...
+    /// Tests if negative index yields an error in get_coeff with [`Zq`].
+    #[test]
+    #[should_panic]
+    fn negative_index_error_zq() {
+        let poly = ModulusPolynomialRingZq::from_str("4  0 1 2 3 mod 17").unwrap();
+
+        let _: Zq = poly.get_coeff(-1).unwrap();
+    }
+}
+
```
Marvin-Beckmann, line 237: please add a test for indexes that have not been set yet

Marvin-Beckmann (changes requested):

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
Marcel583: "If an index is provided which exceeds the highest set coefficient, `0` is returned.". This is what we wrote in other implementations too. What should I add here to make it more clear?

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
Marcel583: The first tests ensures this

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
Marcel583: I think that self.get_fq_ctx_struct().ctxp[0] is copied here (Since fmpz_mod_ctx has the Copy trait) and thus, no leak occurs

Marvin-Beckmann: I see... but their copy trait implementation is wrong... It is derived, but it has fmpz inside of it, which are not properly copied, meaning that this object causes memory leaks

Marcel583: OK, that is bad. I resolved it now by creating a new instance based on the reference to the ctx object

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -8,11 +8,124 @@
...
+        unsafe {
+            fmpz_mod_poly_get_coeff_fmpz(
+                &mut out.value,
+                &self.modulus.modulus[0],
+                index,
+                &self.get_fq_ctx_struct().ctxp[0],
+            )
+        }
+
+        Ok(out)
```
Marvin-Beckmann, line 126: wet code -> you have written the same code twice. I would suggest to find a way to reduce the number of lines, e.g. by calling these lines in the other function or vice-versa

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -8,11 +8,124 @@
...
+            );
+
+            fmpz_mod_ctx_init(ctx.as_mut_ptr(), &self.get_fq_ctx_struct().ctxp[0].n[0]);
+
+            let modulus = Modulus {
+                modulus: Rc::new(ctx.assume_init()),
+            };
+
+            Ok(Zq::from((out_z, modulus)))
+        }
```
Marvin-Beckmann, line 77:
```suggestion
        let entry: Z = self.get_entry(index)?;
        let modulus = self.get_q();
        
        Ok(Zq::from((entry, modulus))) 
```

As we are not able to share a reference counter here, the implementation I suggested is equivalent, but better to maintain (as it is shorter)

Marvin-Beckmann (approved):

PR #424, opened 2024-10-09
UTF8-De-/Encoding for MatZ

jnsiemer:
**Description**

This PR implements...
- [x] padding for UTF8-encoded strings to ensure that no problems happen along the way of decoding (including major progress for MatZq already)
- [x] from_utf8
- [x] to_utf 

for `MatZ`.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_z/from.rs
+++ b/src/integer/mat_z/from.rs
@@ -14,10 +14,14 @@ use super::MatZ;
 use crate::{
     error::MathError,
     integer::Z,
-    traits::SetEntry,
-    utils::{dimensions::find_matrix_dimensions, parse::parse_matrix_string},
+    traits::{GetNumColumns, GetNumRows, SetEntry},
+    utils::{
+        dimensions::find_matrix_dimensions,
+        index::evaluate_indices,
+        parse::{matrix_from_utf8_fill_bytes, parse_matrix_string},
+    },
 };
-use std::str::FromStr;
+use std::{fmt::Display, str::FromStr};
 
 impl FromStr for MatZ {
     type Err = MathError;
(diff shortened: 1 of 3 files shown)


src/integer/mat_z/to_string.rs
```diff
@@ -144,3 +189,46 @@ mod test_to_string {
...
+    /// if the integer contains an invalid UTF8-Encoding.
+    #[test]
+    fn invalid_encoding() {
+        // 128 is an invalid UTF8-character (at least at the end and on its own)
+        let matrix = MatZ::from_str("[[1,2],[3,128]]").unwrap();
+        let string = matrix.to_utf8();
+
+        assert!(string.is_err());
+    }
+}
```
Marvin-Beckmann, line 234: From what I can see some unintended behaviour can occur.

I know that we only want to use it by:
1. Create a string
2. Create a matrix

But the other direction should not be forgotten, because the other behaviour might also be worth considering:
1. Create a matrix
2. Create a string
-> here we should run into an issue: The roundtrip only works if all entries take the same number of bits, meaning that if you have an unbalanced matrix and call to_utf8, this might be unexpected from the perspective of a user if not documented well. Also I am not sure about the behaviour in general.

Also: as you are implementing it for several types and have the same function signature it might make sense to introduce a trait for this functionality

src/integer/mat_z/to_string.rs
```diff
@@ -67,6 +72,46 @@ impl fmt::Display for MatZ {
     }
 }
 
+impl MatZ {
+    /// Enables conversion to a UTF8-Encoded [`String`] for [`MatZ`] values.
+    /// The inverse to this function is [`MatZ::from_utf8`] for valid UTF8-Encodings.
+    ///
+    /// **Warning**: Not every byte-sequence forms a valid UTF8-character.
+    /// If this is the case, a [`FromUtf8Error`] will be returned.
```
Marvin-Beckmann, line 80: hmm, I am not sure if this is really helpful to a user

Marvin-Beckmann (commented):
Hi, I just provided some thoughts while looking over the PR. Not sure when I find time to do a full review, but here are some of my thoughts...

src/integer/mat_z/from.rs
```diff
@@ -76,6 +80,56 @@ impl From<&MatZ> for MatZ {
...
+        num_cols: impl TryInto<i64> + Display,
+    ) -> Self {
+        let (num_rows_i64, num_cols_i64) = evaluate_indices(num_rows, num_cols).unwrap();
+        let mut mat = MatZ::new(num_rows_i64, num_cols_i64);
+        let num_columns = mat.get_num_columns() as usize;
+        let nr_entries = mat.get_num_rows() as usize * num_columns;
+
+        // This error can't be triggered as no modulus is provided.
+        let (byte_vector, nr_bytes_per_entry) =
+            matrix_from_utf8_fill_bytes(message, nr_entries, None).unwrap();
```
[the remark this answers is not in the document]
jnsiemer: It's being padded until the same number of bytes is available for each entry.

src/integer/mat_z/to_string.rs
jnsiemer: This shouldn't be an issue. Only the `from_utf8` function works with entries that need to have the same number of bytes for each entry and it applies padding before. The `to_utf8` function doesn't require a constant number of bytes per entry. it just takes the available bytes and appends them to the existing vector. If that is a valid UTF8-Sequence, great! Otherwise, that's a problem of the developer. In this case, I don't see a reason to introduce any limitations.

Marvin-Beckmann: I agree that the functionality itself works, but I just wanted to say that inconsistencies might occur because of the implementation...

I want a test that instantiates a matrix with a large entry (like 128 bits or something) and one entry with a small number of bits. Then, calling to_utf8 and afterwards calling from_utf8 will most likely give you a different matrix than the original one

Marvin-Beckmann (commented):
@Marcel583 Can you do the full review? I just don't find the time for it...

src/utils/parse.rs
```diff
@@ -245,3 +307,157 @@ mod test_matrix_to_string {
...
+                    0.0f32,
+                    (byte_vector.len() as f32 / matrix_size as f32).fract()
+                );
+                assert_eq!(48u8, byte_vector[byte_vector.len() - 1]);
+            }
+        }
+    }
+
+    /// Ensures that a no bytes are added if it isn't necessary, i.e.
+    /// if every entry is naturally allocated the same number of bytes.
```
Marcel583, line 351:
```suggestion
    /// Ensures that no bytes are added if it isn't necessary, i.e.
    /// if every entry is naturally allocated the same number of bytes.
```

src/utils/parse.rs
```diff
@@ -245,3 +307,157 @@ mod test_matrix_to_string {
...
+        let matrix_size = 2;
+        let cmp_vec: Vec<u8> = vec![];
+
+        let (byte_vector, nr_bytes_per_entry) =
+            matrix_from_utf8_fill_bytes(message, matrix_size, None).unwrap();
+
+        assert_eq!(cmp_vec, byte_vector);
+        assert_eq!(0, nr_bytes_per_entry);
+    }
+
```
Marcel583, line 453: Please add a test where the ConversionError is triggered.

Marcel583 (approved):
Looks great. Just some things in the doc comments I would change and a test case I would add.

PR #426, opened 2024-10-25
From for ModulusPolyRingZq

Marcel583:
**Description**

This PR corrects the current implementation of ModulusPolyRingZq, since it did not check the polynomial modulus to be unequal to 0 and 1. 
It also implements...
- [x] From (&)Zq
- [x] From (&)PolyOverZ, Mod
- [x] From Integer, Mod

for ModulusPolyRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
@@ -331,12 +331,13 @@ mod test_try_from_poly_z {
     use crate::{integer::PolyOverZ, integer_mod_q::ModulusPolynomialRingZq};
     use std::str::FromStr;
 
-    /// Ensure that non-primes work
+    /// Ensure that primes and non-primes work as modulus
     #[test]
-    fn poly_z_non_prime() {
+    fn poly_z_primes() {
         let poly_z = PolyOverZ::from_str("2  2 2").unwrap();
 
-        let _ = ModulusPolynomialRingZq::from((poly_z, 10));
+        let _ = ModulusPolynomialRingZq::from((&poly_z, 10));
+        let _ = ModulusPolynomialRingZq::from((poly_z, 11));
     }
 
     /// Ensure that the function panics if the modulus polynomial is 0
@@ -364,10 +365,11 @@ mod test_try_from_poly_z {
 mod test_try_from_integer_mod {
     use crate::integer_mod_q::ModulusPolynomialRingZq;
 
-    /// Ensure that non-primes work
+    /// Ensure that primes and non-primes work as modulus
     #[test]
-    fn mod_non_prime() {
+    fn mod_primes() {
         let _ = ModulusPolynomialRingZq::from((5, 10));
+        let _ = ModulusPolynomialRingZq::from((5, 11));
     }
 
     /// Ensure that the function panics if the modulus polynomial is 0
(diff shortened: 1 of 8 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -11,16 +11,155 @@
...
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::{integer_mod_q::*, traits::*};
+    ///
+    /// let mod_poly = ModulusPolynomialRingZq::from(&Zq::from((2, 10)));
+    ///
+    /// let poly_cmp = ModulusPolynomialRingZq::from((2, 10));
+    /// assert_eq!(mod_poly, poly_cmp);
+    /// assert_eq!(mod_poly.get_degree(), 0);
```
jnsiemer, line 41:
```suggestion
    /// use qfall_math::{integer_mod_q::*, traits::*};
    ///
    /// let zq = Zq::from((2, 10));
    ///
    /// let mod_poly = ModulusPolynomialRingZq::from(&zq);
```
Just a suggestion. Looks more concise to me as an example on how to use the function.

src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -89,15 +234,156 @@ impl FromStr for ModulusPolynomialRingZq {
...
+/// let poly_zq = PolyOverZq::from_str("2  1 2 mod 17").unwrap();
+///
+/// check_poly_mod(&poly_zq)?
+/// ```
+///
+/// # Errors and Failures
+/// - Returns a [`MathError`] of type
+///     [`InvalidModulus`](MathError::InvalidModulus)
+///     if the modulus polynomial is `0` or `1`.
+pub(crate) fn check_poly_mod(poly_zq: &PolyOverZq) -> Result<(), MathError> {
```
jnsiemer, line 281: Why didn't you just use a `PolyOverZ` as input? You don't check the modulus anyway.

Furthermore, I don't know how a function behaves that has a negative modulus-polynomial like -1. I would believe that this wouldn't allow to host any information as well (just like 1 wouldn't allow it). Please speak to Laurens to check for all possible cases in this method before you merge it.

src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
```diff
@@ -89,15 +234,156 @@ impl FromStr for ModulusPolynomialRingZq {
...
+    use crate::{integer::PolyOverZ, integer_mod_q::ModulusPolynomialRingZq};
+    use std::str::FromStr;
+
+    /// Ensure that non-primes work
+    #[test]
+    fn poly_z_non_prime() {
+        let poly_z = PolyOverZ::from_str("2  2 2").unwrap();
+
+        let _ = ModulusPolynomialRingZq::from((poly_z, 10));
+    }
```
jnsiemer, line 342: If you add a test for a non-prime, why isn't there one for primes?
I don't think that you need to introduce two different tests. I believe that these tests should be renamed and their description updated to be more general.

jnsiemer (approved):
Great work! Please consider changes according to my comments before merging.

src/integer_mod_q/modulus_polynomial_ring_zq/from.rs
Marcel583: We need a PolyOverZq for the conversion into a ModulusPolynomialRingZq and since the function is only used for conversions to that type, it is faster to insert the PolyOverZq into the function rather than transforming it into a PolyOverZ.

I spoke with Laurens about that and we agreed on only rejecting 0 polynomials and allowing both 1 and -1.

PR #431, opened 2024-11-15
MatPolyRingZq MismatchingModulus error for concatenation

Marcel583:
**Description**

This PR implements the MismatchingModulus error for concatenation of MatPolyRingZq elements.

**Testing**

**Checklist:**


Marvin-Beckmann (approved):
good catch. Next time: please comment the exact bug that was found in the comment of the PR, then one does not have to find out what was wrong themselves (even though it was clear here)

PR #434, opened 2024-11-20
PolyOverZq, ModulusPoly get_representative

Marcel583:
**Description**

This PR implements get_representative for PolyOverZq and ModulusPolynomialRingZq. Furthermore, it deletes from<PolyOverZq> for PolyOverZ and renames the get_representative functions to a more meaningful name.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
@@ -10,8 +10,8 @@
 
 use crate::{
     error::MathError,
-    integer::Z,
-    integer_mod_q::{Modulus, ModulusPolynomialRingZq, Zq},
+    integer::{PolyOverZ, Z},
+    integer_mod_q::{Modulus, ModulusPolynomialRingZq, PolyOverZq, Zq},
     traits::GetCoefficient,
     utils::index::evaluate_index,
 };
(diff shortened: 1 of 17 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
```diff
@@ -163,6 +163,28 @@ impl ModulusPolynomialRingZq {
...
+    ///
+    /// let poly_z = modulus.get_representative_0_modulus();
+    ///
+    /// let cmp_poly = PolyOverZ::from_str("4  14 0 14 1").unwrap();
+    /// assert_eq!(cmp_poly, poly_z);
+    /// ```
+    pub fn get_representative_0_modulus(&self) -> PolyOverZ {
+        let poly_zq = PolyOverZq::from(self);
+        poly_zq.get_representative_0_modulus()
+    }
```
Marvin-Beckmann, line 187: what do you think about the name:
- ``get_representative_least_positive_residue``

This seems to be more of a standardised name and at first I was also somewhat confused what was meant by the current name. What do you think @jnsiemer ?

Marvin-Beckmann (changes requested):
I have just seen that the naming was already changed in a previous commit, but I have a different suggestion that would be more speaking in terms of its functionality

The content of the PR seems fine otherwise

Please do not forget subsequent changes in the math crate afterwards

src/integer_mod_q/modulus_polynomial_ring_zq/get.rs
Marvin-Beckmann: One could also rename the other function to ``get_representative_least_absolute_residue`` to be consistent if you like the suggestion. Although I do not find the other functions name to be as misleading

Marvin-Beckmann (approved):

PR #442, opened 2025-01-06
ModPoly coefficient embedding

Marcel583:
**Description**

This PR implements coefficient embedding for ModulusPolynomialRingZq according to the implementation for PolyOverZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/coefficient_embedding.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/coefficient_embedding.rs
@@ -1,0 +1,181 @@
+// Copyright © 2025 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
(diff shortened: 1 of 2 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,181 @@
...
+        let mut out = MatZq::new(size, 1, self.get_q());
+        for j in 0..size {
+            let coeff: Result<Z, _> = self.get_coeff(j);
+            match coeff {
+                Ok(value) => out.set_entry(j, 0, value).unwrap(),
+                Err(_) => break,
+            }
+        }
+        out
+    }
```
Marvin-Beckmann, line 65: I think we should rather implement a method to return the PolyOverZq that identifies the object with a dedicated function and just call its function to provide this functionality.
Specifically, we have a function to get the `q`, but not the polynomial, so I think that should be done first

Marvin-Beckmann (changes requested):

PR #445, opened 2025-01-12
PartialEq between Z and Q

Marcel583:
**Description**

This PR implements PartialEq for all combinations of borrowed and owned values of Q, Z and other integer values like fmpz and i8. 

As for arithmetics I have implemented PartialEq<Z> for Q and PartialEq<Q> for Z. The implementation for combinations of borrowed and owned values is implemented via a macro in the same style as for arithmetics. The implementation for other integer values uses the for_others macro that uses Z as a bridge type.

There are still some clippy warnings that will be gone after I implement the same for PartialOrd. The marco module cmp.rs will also contain a macro for PartialOrd, therefore the doc comment for the module already includes PartialOrd.

**Testing**

**Checklist:**


src/integer_mod_q/polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+        out
+    }
+}
+
+impl FromCoefficientEmbedding<&MatZq> for PolynomialRingZq {
+    /// Computes a polynomial from a vector.
+    /// The first i-th entry of the column vector is taken
+    /// as the coefficient of the polynomial.
```
jnsiemer, line 74:
```suggestion
    /// Computes a polynomial of degree `n-1` from a column vector of size `n`.
    /// The i-th entry of the column vector is taken
    /// as the i-th coefficient of the polynomial.
```

src/integer_mod_q/polynomial_ring_zq/evaluate.rs
```diff
@@ -0,0 +1,255 @@
...
+#[cfg(test)]
+mod test_evaluate_z {
+    use crate::integer::Z;
+    use crate::integer_mod_q::{PolyOverZq, Zq};
+    use crate::traits::Evaluate;
+    use std::str::FromStr;
+
+    /// Tests if evaluate works for [`Z`] as input
+    #[test]
+    fn eval_z() {
```
jnsiemer, line 126: I would also introduce a test where evaluated value is large enough to cross the modulus, i.e. to check if the modulus is applied (although this is partly done by the negative test, yes).

jnsiemer (approved):
Overall great and a lot of work!
There are several small impurities that I've marked. Some of them should be addressed, others don't necessarily need to. However, I'm sure you'll resolve them with closed eyes.

PR #449, opened 2025-02-08
PolynomialRingZq coefficient embedding

Marcel583:
**Description**

This PR implements the coefficient embedding for PolynomialRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_poly_over_z/coefficient_embedding.rs
+++ b/src/integer/mat_poly_over_z/coefficient_embedding.rs
@@ -146,7 +146,9 @@ impl IntoCoefficientEmbedding<MatZ> for &MatPolyOverZ {
     ///     not all coefficients can be embedded.
     /// - if `self` is not a row vector
     fn into_coefficient_embedding(self, size: impl Into<i64>) -> MatZ {
-        assert!(self.is_row_vector());
+        assert!(self.is_row_vector(),
+        "This is no valid input, since the matrix is no row vector."
+    );
         let size = size.into();
 
         let mut poly: Vec<PolyOverZ> = Vec::new();
--- a/src/integer/mat_poly_over_z/coefficient_embedding.rs
+++ b/src/integer/mat_poly_over_z/coefficient_embedding.rs
@@ -11,7 +11,7 @@
 
 use super::MatPolyOverZ;
 use crate::{
-    integer::{MatZ, PolyOverZ},
+    integer::{MatZ, PolyOverZ, Z},
     traits::{
         Concatenate, FromCoefficientEmbedding, GetCoefficient, GetEntry, GetNumColumns, GetNumRows,
         IntoCoefficientEmbedding, SetEntry,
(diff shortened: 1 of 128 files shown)


src/integer_mod_q/poly_over_zq/coefficient_embedding.rs
```diff
@@ -97,7 +97,10 @@ impl FromCoefficientEmbedding<&MatZq> for PolyOverZq {
     /// # Panics ...
     /// - if the provided embedding is not a column vector.
     fn from_coefficient_embedding(embedding: &MatZq) -> Self {
-        assert!(embedding.is_column_vector());
+        assert!(
+            embedding.is_column_vector(),
+            "This is no valid embedding, since the matrix is no column vector."
+        );
```
Marvin-Beckmann, line 103: If a panic occurs, it also shows where the function was called, i.e., which call was the reason. I think, that the additional comment does not provide further explanation than the code, so I would not include the additional comment. If you strongly want to keep it - I am also fine with that

src/integer_mod_q/polynomial_ring_zq/coefficient_embedding.rs
Marvin-Beckmann, line 1: I have to think more about the purpose of this part, and if it itself makes sense as an implementation due to non-unique representations

src/integer/mat_poly_over_z/coefficient_embedding.rs
```diff
@@ -146,7 +146,10 @@ impl IntoCoefficientEmbedding<MatZ> for &MatPolyOverZ {
     ///     not all coefficients can be embedded.
     /// - if `self` is not a row vector
     fn into_coefficient_embedding(self, size: impl Into<i64>) -> MatZ {
-        assert!(self.is_row_vector());
+        assert!(
+            self.is_row_vector(),
+            "This is no valid input, since the matrix is no row vector."
+        );
```
jnsiemer, line 152: We should consider removing this requirement to open the library up in case someone wants to use the module-setting. - Anywhere this change is made, I could add this comment (and anywhere else in the library)

src/integer_mod_q/polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,236 @@
...
+            the provided size ({size})."
+        );
+        let mut out = MatZq::new(size, 1, self.modulus.get_q());
+        for j in 0..size {
+            let coeff: Result<Z, _> = self.get_coeff(j);
+            match coeff {
+                Ok(value) => out.set_entry(j, 0, value).unwrap(),
+                Err(_) => break,
+            }
+        }
```
jnsiemer, line 69:
```suggestion
        for j in 0..self.get_degree() {
            let coeff: Z = self.get_coeff(j).unwrap();
            out.set_entry(j, 0, value).unwrap();
        }
```
If you iterate over the size of the polynomial, you should be guaranteed to not receive any errors (and be more efficient as you leave out iterations for any 0 entries at the bottom).

src/integer_mod_q/polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// Ensure that the doc test works.
+    #[test]
+    fn test_example() {
+        let poly = PolynomialRingZq::from_str("2  1 -2 / 3  17 3 5 mod 19").unwrap();
+        let embedding = poly.into_coefficient_embedding(3);
+
+        let cmp_vector = MatZq::from_str("[[1],[-2],[0]] mod 19").unwrap();
+        assert_eq!((cmp_vector, poly.get_mod()), embedding);
+    }
+
```
jnsiemer, line 140:
```suggestion
```
You don't need to copy the doc-test here. It's going to be executed as a doc-test anyway. If you want a small_entries test, you can add one, but then it should be a different one. - There is no point in running the same test twice (if there's no randomness involved).

jnsiemer (approved):
Please consider applying the suggestions before merging - great work!

src/integer_mod_q/polynomial_ring_zq/coefficient_embedding.rs
Marcel583: Yes, this is true. It is a copy of our underlying code for all coefficient embeddings. I will change this in all instances.

PR #450, opened 2025-02-08
MatPolyRingZq coefficient embedding

Marcel583:
**Description**

This PR implements coefficient embedding for MatPolyRingZq.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
+++ b/src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
@@ -1,0 +1,501 @@
+// Copyright © 2025 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! This module contains implementations to transform a [`MatPolynomialRingZq`]
+//! into a [`MatZq`] and a [`ModulusPolynomialRingZq`] and reverse by using the coefficient embedding.
+
+use super::MatPolynomialRingZq;
+use crate::{
+    integer::Z,
+    integer_mod_q::{MatZq, ModulusPolynomialRingZq, PolynomialRingZq, Zq},
+    traits::{
+        Concatenate, FromCoefficientEmbedding, GetCoefficient, GetEntry, GetNumColumns, GetNumRows,
+        IntoCoefficientEmbedding, SetEntry,
+    },
+};
+
+impl MatPolynomialRingZq {
+    /// Computes the coefficient embedding of a matrix of polynomials.
+    /// All rows are independently embedded using the coefficient embedding
+    /// and then vertically concatenated to an embedding.
+    /// It inverts the operation of
+    /// [`MatPolynomialRingZq::from_coefficient_embedding_to_matrix`].
+    ///
+    /// Parameters:
+    /// - `size`: determines the number of rows each polynomial is embedded in.
+    ///     It has to be larger than the degree of all polynomials.
+    ///
+    /// Returns a coefficient embedding as a matrix if `size` is large enough.
+    ///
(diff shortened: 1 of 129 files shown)


src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,501 @@
...
+    ///     not all coefficients can be embedded.
+    pub fn into_coefficient_embedding_from_matrix(
+        &self,
+        size: impl Into<i64>,
+    ) -> (MatZq, ModulusPolynomialRingZq) {
+        let size = size.into();
+        let mut embedded = self.get_row(0).unwrap().into_coefficient_embedding(size);
+        for i in 1..self.get_num_rows() {
+            let self_i = self.get_row(i).unwrap().into_coefficient_embedding(size);
+            embedded.0 = embedded.0.concat_vertical(&self_i.0).unwrap();
```
jnsiemer, line 59: Could it be more efficient to create a matrix of appropriate size at the beginning and then modify its entries afterwards? As this action might be called several times during the execution of a scheme, execution time might be relevant at this point and memory management should theoretically be more efficient this way.

src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,501 @@
...
+                .unwrap();
+            row_poly_mat.push(MatPolynomialRingZq::from_coefficient_embedding((
+                &sub_mat_i,
+                embedding.1,
+            )));
+        }
+
+        let mut out = row_poly_mat.first().unwrap().clone();
+        for poly_vec in row_poly_mat.iter().skip(1) {
+            out = out.concat_vertical(poly_vec).unwrap();
```
jnsiemer, line 130: Same question at this point: Is it more efficient to create a matrix of the correct at the beginning?

jnsiemer (approved):
Please check my two comments reg. efficiency before merging

src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// let poly = MatPolynomialRingZq::from_str("[[1  1, 2  1 2],[1  -1, 2  -1 -2]] / 3  1 2 3 mod 17").unwrap();
+    /// let embedding = poly.into_coefficient_embedding_from_matrix(2);
+    /// let cmp_mat = MatZq::from_str("[[1, 1],[0, 2],[-1, -1],[0, -2]] mod 17").unwrap();
+    /// assert_eq!((cmp_mat, poly.get_mod()), embedding);
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if `size` is not larger than the degree of the polynomial, i.e.
+    ///     not all coefficients can be embedded.
+    pub fn into_coefficient_embedding_from_matrix(
```
jnsiemer, line 51: We should probably reconsider this name - "into_coefficient_embedding" should suffice. There shouldn't be a difference if it's a row-vector or a matrix - one is the ring-setting, the other the module-setting. Same thing for both. So, why not pack this up in one function with "into_coefficient_embedding"?

src/integer_mod_q/mat_polynomial_ring_zq/coefficient_embedding.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// assert_eq!((cmp_mat, poly.get_mod()), embedding);
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if `size` is not larger than the degree of the polynomial, i.e.
+    ///     not all coefficients can be embedded.
+    /// - if `self` is not a row vector
+    fn into_coefficient_embedding(self, size: impl Into<i64>) -> (MatZq, ModulusPolynomialRingZq) {
+        assert!(
+            self.is_row_vector(),
```
jnsiemer, line 169: We should get rid of this to enable this function for the module-setting. It looks like an unnecessary requirement to me.

PR #452, opened 2025-02-09
Is_symmetric for all matrices

Marcel583:
**Description**

This PR implements is_symmetric for MatZ, MatZq, MatPolyOverZ and MatPolynomialRingZq, based on the implementation for MatQ.

**Testing**

**Checklist:**


src/integer/mat_poly_over_z/properties.rs
```diff
@@ -73,6 +73,21 @@ impl MatPolyOverZ {
...
+    ///
+    /// # Examples
+    /// ```
+    /// use qfall_math::integer::MatPolyOverZ;
+    ///
+    /// let value = MatPolyOverZ::identity(2,2);
+    /// assert!(value.is_symmetric());
+    /// ```
+    pub fn is_symmetric(&self) -> bool {
+        self == &self.transpose()
```
jnsiemer, line 88: Please check whether the comparison via "self.get_entry(i,j) == self.get_entry(n-i, n-j)" isn't more efficient. The efficiency is probably not required to be the most efficient, but it'd be intersting to know how they compare on a larger scale. Thanks!

jnsiemer (approved):
Please consider my comment reg. efficiency before merging. Thanks!

PR #453, opened 2025-02-13
Revise `sample_z`

jnsiemer:
**Description**

This PR implements `DiscreteGaussianIntegerSampler` and sample_z to
- [x] store evaluations of the Gaussian function efficiently (requires implementation of `Hash` for `Z`),
- [x] use f64::exp instead of Q::exp (roughly twice as fast - though there is no specification given for its precision),
- [x] corrects the interval that sample_z samples uniformly from, which was one entry too small,
- [x] benchmarks for sample_z.

Overall, the new implementation reduces discrete Gaussian sampling for a single sample by roughly 10 µs for the evaluated parameters (from 28 µs to 17 µs for a wide distribution and 18 µs to 9 µs for a narrow distribution).
If we sample multiple samples from the same distribution (usually done to sample the G-trapdoor R or preimages for G), then the speedup is just shy of 90% for 10,000 samples (280 ms -> 32 ms for a wide and 180 ms -> 19 ms for a narrow distribution).

**Testing**

**Checklist:**

### DIFF
--- a/benches/sample_z.rs
+++ b/benches/sample_z.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Niklas Siemer
+// Copyright © 2025 Niklas Siemer
 //
 // This file is part of qFALL-math.
 //
--- a/benches/sample_z.rs
+++ b/benches/sample_z.rs
@@ -6,7 +6,7 @@
 // the terms of the Mozilla Public License Version 2.0 as published by the
 // Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
 
-//! Create benchmark for integers in this file.
+//! Create benchmarks for `sample_z` and discrete Gaussian sampling in this file.
 
 use criterion::*;
 use qfall_math::{
@@ -14,59 +14,59 @@ use qfall_math::{
     rational::Q,
 };
 
-/// Create a matrix of size 100x100 sampled by a comparatively wide discrete Gaussian distribution.
-pub fn sample_z_wide() {
-    let n = Z::from(1000);
-    let c = Q::from(0);
-    let s = Q::from(100);
-
-    let _ = MatZ::sample_discrete_gauss(100, 100, &n, &c, &s).unwrap();
-}
-
-/// Create a matrix of size 100x100 sampled by a comparatively narrow discrete Gaussian distribution.
-pub fn sample_z_narrow() {
-    let n = Z::from(100);
-    let c = Q::from(0);
-    let s = Q::from(2);
-
-    let _ = MatZ::sample_discrete_gauss(100, 100, &n, &c, &s).unwrap();
-}
-
-/// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
-pub fn sample_z_wide_single() {
-    let n = Z::from(1000);
-    let c = Q::from(0);
-    let s = Q::from(100);
-
-    let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
-}
-
-/// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
-pub fn sample_z_narrow_single() {
-    let n = Z::from(100);
-    let c = Q::from(0);
-    let s = Q::from(2);
-
-    let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
-}
-
-/// benchmark [sample_z_wide]
+/// benchmark creating a matrix of size 100x100 sampled by a comparatively wide discrete Gaussian distribution.
 pub fn bench_sample_z_wide(c: &mut Criterion) {
+    /// Create a matrix of size 100x100 sampled by a comparatively wide discrete Gaussian distribution.
+    fn sample_z_wide() {
+        let n = Z::from(1000);
+        let center = Q::from(0);
+        let s = Q::from(100);
+
+        let _ = MatZ::sample_discrete_gauss(100, 100, n, center, s).unwrap();
+    }
+
     c.bench_function("SampleZ wide 10,000", |b| b.iter(sample_z_wide));
 }
 
-/// benchmark [sample_z_narrow]
+/// benchmark creating a matrix of size 100x100 sampled by a comparatively narrow discrete Gaussian distribution.
 pub fn bench_sample_z_narrow(c: &mut Criterion) {
+    /// Create a matrix of size 100x100 sampled by a comparatively narrow discrete Gaussian distribution.
+    fn sample_z_narrow() {
+        let n = Z::from(100);
+        let c = Q::from(0);
+        let s = Q::from(2);
+
+        let _ = MatZ::sample_discrete_gauss(100, 100, &n, &c, &s).unwrap();
+    }
+
     c.bench_function("SampleZ narrow 10,000", |b| b.iter(sample_z_narrow));
 }
 
-/// benchmark [sample_z_wide_single]
+/// benchmark creating a single integer sampled by a comparatively wide discrete Gaussian distribution.
 pub fn bench_sample_z_wide_single(c: &mut Criterion) {
+    /// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
+    fn sample_z_wide_single() {
+        let n = Z::from(1000);
+        let c = Q::from(0);
+        let s = Q::from(100);
+
+        let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
+    }
+
     c.bench_function("SampleZ wide single", |b| b.iter(sample_z_wide_single));
 }
 
-/// benchmark [sample_z_narrow_single]
+/// benchmark creating a single integer sampled by a comparatively wide discrete Gaussian distribution.
 pub fn bench_sample_z_narrow_single(c: &mut Criterion) {
+    /// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
+    fn sample_z_narrow_single() {
+        let n = Z::from(100);
+        let c = Q::from(0);
+        let s = Q::from(2);
+
+        let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
+    }
+
     c.bench_function("SampleZ narrow single", |b| b.iter(sample_z_narrow_single));
 }
 
--- a/benches/sample_z.rs
+++ b/benches/sample_z.rs
@@ -1,0 +1,79 @@
+// Copyright © 2023 Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Create benchmark for integers in this file.
+
+use criterion::*;
+use qfall_math::{
+    integer::{MatZ, Z},
+    rational::Q,
+};
+
+/// Create a matrix of size 100x100 sampled by a comparatively wide discrete Gaussian distribution.
+pub fn sample_z_wide() {
+    let n = Z::from(1000);
+    let c = Q::from(0);
+    let s = Q::from(100);
+
+    let _ = MatZ::sample_discrete_gauss(100, 100, &n, &c, &s).unwrap();
+}
+
+/// Create a matrix of size 100x100 sampled by a comparatively narrow discrete Gaussian distribution.
+pub fn sample_z_narrow() {
+    let n = Z::from(100);
+    let c = Q::from(0);
+    let s = Q::from(2);
+
+    let _ = MatZ::sample_discrete_gauss(100, 100, &n, &c, &s).unwrap();
+}
+
+/// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
+pub fn sample_z_wide_single() {
+    let n = Z::from(1000);
+    let c = Q::from(0);
+    let s = Q::from(100);
+
+    let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
+}
+
+/// Create a single integer sampled by a comparatively wide discrete Gaussian distribution.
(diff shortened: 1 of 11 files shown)


src/integer/z/hash.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright © 2023 Niklas Siemer
```
Marvin-Beckmann, line 1:
```suggestion
// Copyright © 2025 Niklas Siemer
```

src/integer/z/hash.rs
```diff
@@ -0,0 +1,29 @@
...
+impl Hash for Z {
+    /// Ensure that the [`flint_sys::fmpz::fmpz`] value is below the
+    /// threshold of being a pointer to the memory using modulus.
+    ///
+    /// Otherwise, this function uses the hash of [`i64`].
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let modulo_value = self.modulo(MODULUS).value.0;
+        modulo_value.hash(state);
+    }
+}
```
Marvin-Beckmann, line 29: I think this is not really intuitive.

Our Z's are large integers, and only being able to hash i64's is kind of strange.
Admittedly, you do allow for higher values, but it is easy to find hash collisions here.
This does not really seem nice.

I see the incentive to do it only over i64, because it is faster, but given the non-uniqueness, this does not seem ideal to me...
I would prefer to work with unique representations(e.g. bitstrings of the actual values) instead of adding additional hash collisions.


For our purposes here these collisions will likely never happen, specifically because they are very far apart ,...

However, if this is aimed to be used by others, this might cause problems

src/integer/z/hash.rs
```diff
@@ -0,0 +1,29 @@
...
+//! `Z` is a type for integers with arbitrary length.
+//! This implementation uses the [FLINT](https://flintlib.org/) library.
+
+use super::Z;
+use std::hash::Hash;
+
+// The modulus is chosen as large as possible, i.e. slighly below 2^62.
+// Furthermore, to ensure that multiplicative actions result in changes with
+// high probability, optimizing when it is prime, i.e. behaves like a field.
+static MODULUS: u64 = 4611686018427387847; // 2^62 - 57 is the largest prime below 2^62
```
Marvin-Beckmann, line 18: I do not really get this, why exactly prime?
But maybe that is just because I think that the object itself is strange

src/utils/sample/discrete_gauss.rs
```diff
@@ -24,23 +24,18 @@ use crate::{
     traits::{GetNumColumns, GetNumRows, Pow},
 };
 use rand::RngCore;
+use std::collections::HashMap;
 
-/// Chooses a sample according to the discrete Gaussian distribution out of
-/// `[center - ⌈s * log_2(n)⌉ , center + ⌊s * log_2(n)⌋ ]`.
+/// Enables for discrete Gaussian sampling out of
+/// `[⌈center - s * log_2(n)⌉ , ⌊center + s * log_2(n)⌋ ]`.
+/// This struct computes the Gauss function lazily and saves it in a [`HashMap`].
```
Marvin-Beckmann, line 31: At first, I was not quite sure what this meant, i.e. if this means that the samples are precomputed or what else it means.
Therefore, I think this could be made a bit more precise

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
+    lower_bound: Z,
+    interval_size: Z,
+    table: HashMap<Z, f64>,
+}
 
-    // sample x in [c - s * log_2(n), c + s * log_2(n)]
-    let mut sample = &lower_bound + sample_uniform_rejection(&interval_size).unwrap();
+impl DiscreteGaussianIntegerSampler {
+    /// Initializes the [`DiscreteGaussianIntegerSampler`] with
+    /// - `center` as the peak of the Gaussian function,
```
Marvin-Beckmann, line 61: yes, but given that we do discrete Gaussian sampling and the center need not be within the lattice, this might sound confusing.
I would just say that it describes the center of the discrete Gaussian to sample from

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
+    /// - `s` defining the Gaussian parameter, which is proportional
+    ///   to the standard deviation `sigma * sqrt(2 * pi) = s`,
+    /// - `lower_bound` as `⌈center - s * log_2(n)⌉`,
+    /// - `interval_size` as `⌊center + s * log_2(n)⌋ - ⌈center - s * log_2(n)⌉`, and
+    /// - `table` as an empty [`HashMap`] to store evaluations of the Gaussian function in.
```
Marvin-Beckmann, line 66:
```suggestion
    /// - `table` as an empty [`HashMap`] to store evaluations of the Gaussian function.
```

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
+    ///
+    /// Parameters:
+    /// - `n`: specifies the range from which is sampled
+    /// - `center`: specifies the position of the center with peak probability
```
Marvin-Beckmann, line 70: see above

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
+                This function expects this input to be larger than 0."
+            )));
+        }
+        if s * n.log(2).unwrap() < Q::ONE {
+            return Err(MathError::InvalidIntegerInput(format!(
+                "The size {s} * log_2({n}) is smaller than 1.
+                Hence, the interval to sample from is of size 1.
+                Please provide larger parameters to sample discrete Gaussian."
+            )));
+        }
```
Marvin-Beckmann, line 111: I agree that it does not really makes sense to sample here, but should we prohibit it?

I am fine either way, just know, that we would have to add this everywhere

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
 
-    // sample x in [c - s * log_2(n), c + s * log_2(n)]
-    let mut sample = &lower_bound + sample_uniform_rejection(&interval_size).unwrap();
+impl DiscreteGaussianIntegerSampler {
+    /// Initializes the [`DiscreteGaussianIntegerSampler`] with
+    /// - `center` as the peak of the Gaussian function,
+    /// - `s` defining the Gaussian parameter, which is proportional
+    ///   to the standard deviation `sigma * sqrt(2 * pi) = s`,
+    /// - `lower_bound` as `⌈center - s * log_2(n)⌉`,
+    /// - `interval_size` as `⌊center + s * log_2(n)⌋ - ⌈center - s * log_2(n)⌉`, and
```
Marvin-Beckmann, line 65: this is not conform with the definition in the code

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
-    // https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=d9f54077d568784c786f7b1d030b00493eb3ae35
-    // this eprint version explains in more detail how sample_z works
-    let mut rng = rand::rng();
-    while gaussian_function(&sample, center, s) <= Q::from((rng.next_u64(), u64::MAX)) {
-        sample = &lower_bound + sample_uniform_rejection(&interval_size).unwrap();
+        let lower_bound = (center - s * n.log(2).unwrap()).ceil();
+        let upper_bound = (center + s * n.log(2).unwrap()).floor();
+        // interval [lower_bound, upper_bound] has upper_bound - lower_bound + 1 elements in it
+        // as uniform sampling samples from [0, interval_size), we require interval_size + 1
+        let interval_size = &upper_bound - &lower_bound + Z::from(2);
```
Marvin-Beckmann, line 117: I think it is only plus 1, because even though uniform sampling samples from [0, interval_size), it samples from an interval that still has size size: `interval_size`, so it should only be +1 imo

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +44,131 @@ use rand::RngCore;
...
+                }
+            } else {
+                // if the function wasn't evaluated for this value before, compute the value and
+                // sample between [0,1] and accept with the value of the Gauss function
+                let evaluated_gauss_function = gaussian_function(&sample, &self.center, &self.s);
+                self.table.insert(sample.clone(), evaluated_gauss_function);
+                if evaluated_gauss_function >= (rng.next_u64() as f64 / u64::MAX as f64) {
+                    return sample;
+                }
+            }
```
Marvin-Beckmann, line 169: I was not able to align it properly, but I think you get what I want

```
            let evaluated_gauss_function = match evaluated_gauss_function {
                Some(x) => x,
                _ => {
                    let evaluated_gauss_function = gaussian_function(&sample, &self.center, &self.s);
                    self.table.insert(sample.clone(), evaluated_gauss_function);
                    evaluated_gauss_function
                }
            };
            if evaluated_gauss_function >= (rng.next_u64() as f64 / u64::MAX as f64) {
                return sample;
```

Marvin-Beckmann: even better than a match might be an unwrap_or_else or something

Marvin-Beckmann (changes requested):
I like the table management to keep Gaussian evaluations, but there are a few things I would like to see addressed

benches/sample_z.rs
```diff
@@ -0,0 +1,79 @@
...
+    let c = Q::from(0);
+    let s = Q::from(2);
+
+    let _ = Z::sample_discrete_gauss(&n, &c, &s).unwrap();
+}
+
+/// benchmark [sample_z_wide]
+pub fn bench_sample_z_wide(c: &mut Criterion) {
+    c.bench_function("SampleZ wide 10,000", |b| b.iter(sample_z_wide));
+}
```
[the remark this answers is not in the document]
jnsiemer: The function in b.iter() needs to be of type `FnMut` and requires to have 0 arguments. There might be a solution to this, but for now this change should be good enough.

src/integer/z/hash.rs
jnsiemer: I'm now using the decimal representation of the value as a String and then the Hash of that String. It should be less efficient than my solution, but meets your requirement to not introduce additional collisions - although having a collision in 2^62 should've been less likely than just filling the RAM entirely.

src/integer/z/hash.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright © 2025 Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! `Z` is a type for integers with arbitrary length.
+//! This implementation uses the [FLINT](https://flintlib.org/) library.
```
Marvin-Beckmann, line 10: make a description related to content

src/utils/sample/discrete_gauss.rs
```diff
@@ -49,42 +45,126 @@ use rand::RngCore;
...
-    let interval_size = &upper_bound - &lower_bound;
+/// let sample = dgis.sample_z();
+/// ```
+pub(crate) struct DiscreteGaussianIntegerSampler {
+    center: Q,
+    s: Q,
+    lower_bound: Z,
+    interval_size: Z,
+    table: HashMap<Z, f64>,
+}
```
Marvin-Beckmann, line 58: for the future: we could think about making this serializable, so it can be reused

src/integer/z/hash.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use super::Z;
+use std::hash::Hash;
+
+impl Hash for Z {
+    /// This function uses the hash of [`String`] with the decimal representation of `self`.
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let string = self.to_string();
+        string.hash(state);
+    }
```
Marvin-Beckmann, line 20: I definitely prefer it, but I have two questions to be considered here:

- Is the conversion to a string adding on overhead larger than the function evaluation, i.e., can you check if the improvement is still as large as before
- can you think of any alternatives that might improve this if the one before is true? e.g. is there a way to make a bitstring? What we could also consider is using a different base representation here, as the to_string uses decimal representation, but other representations would be better compatible with the string type to reduce the length of the string.
- the alternatives that come to my mind are:
  - to_string_b with a higher base than 10
  - to_bytes and try to make it more efficient
  
If you have given these options some thought, I am happy to accept the PR

Marvin-Beckmann (changes requested):

src/integer/z/hash.rs
jnsiemer: I've checked these options before.
Using any type of string or vector of bytes decreases the performance by 25 to 30% (even for base 62). If we don't use a hash of flint, we need to use a hash for a Rust type. As FLINT 3.0 introduces hashing for reals for the first time, we're currently bound to Rust ones.

The question is: How do I choose my modulus? I expect hashing Z's will only be relevant for use in HashMaps such as the one in DiscreteGaussianIntegerSampler. These intervals shouldn't be any larger than 2^62 anyway - the modulo action just ensures that values larger than 2^62 are brought into the non-pointer version to the memory as an i64. The probability of a collision (Birthday bound) is 2^32 for i64 anyway - we reduce it to 2^31 by choosing the modulus 2^62. By choosing it prime, we additionally minimize the probability that multiplication of values does not accidentally map 1 * x and 3 * x is mapped to x by applying the modulus - i.e. it decreases the risk of accidentally introducing hash collisions.

Thus, I chose the option with the largest prime below 2^62 to optimize performance and keep a risk of 2^31 of a collision, which seemed good enough for the application.

Marvin-Beckmann (approved):

PR #454, opened 2025-02-22
PartialEq between Modulus and Z, Q

Marcel583:
**Description**

This PR implements PartialEq between Modulus and Z, and between Modulus and Q. 
It also introduces a macro for automatic implementations of second directions of commutative operations, such that we no longer have to implement both directions of commutative operations. 

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/q/cmp.rs
+++ b/src/rational/q/cmp.rs
@@ -10,7 +10,11 @@
 //! This uses the traits from [`std::cmp`].
 
 use super::Q;
-use crate::{integer::Z, macros::for_others::implement_for_others};
+use crate::{
+    integer::Z,
+    integer_mod_q::Modulus,
+    macros::for_others::{implement_for_others, implement_trait_reverse},
+};
 use flint_sys::{
     fmpq::{fmpq_cmp, fmpq_equal},
     fmpz::{fmpz, fmpz_equal},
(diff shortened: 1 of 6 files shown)


src/rational/q/cmp.rs
```diff
@@ -85,6 +98,52 @@ impl PartialEq<Z> for Q {
...
+    /// # assert!(compared);
+    /// let compared: bool = (Q::eq(&a, &b));
+    /// # assert!(compared);
+    /// let compared: bool = (Modulus::eq(&b, &a));
+    /// # assert!(compared);
+    /// ```
+    fn eq(&self, other: &Modulus) -> bool {
+        unsafe {
+            (1 == fmpz_equal(&self.value.den, &fmpz(1)))
+                && (1 == fmpz_equal(&self.value.num, &other.modulus.n[0]))
```
jnsiemer, line 140: This function assumes that the denominator is 1. What if `Q` isn't simplified? Or is that a global assumption that it is simplified anytime? If not, please adapt the code or add it explicitly in the doc comment.

jnsiemer (approved):
Please rebase and fix the other `clippy` warning. Once you've dealt with my comment, this PR is good to go!

src/rational/q/cmp.rs
Marcel583: We always assume simplified Q values.

//! For **DEVELOPERS**: Many functions assume that the [`Q`] instances are reduced.
//! To avoid unnecessary checks and reductions, always return canonical/reduced
//! values. The end-user should be unable to obtain a non-reduced value.

PR #460, opened 2025-02-24
PartialOrd between Z, Q and  Modulus

Marcel583:
**Description**

This PR implements .

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus/cmp.rs
+++ b/src/integer_mod_q/modulus/cmp.rs
@@ -16,7 +16,7 @@ use crate::{
     integer::Z,
     macros::for_others::{implement_for_others, implement_trait_reverse},
 };
-use flint_sys::fmpz::{fmpz, fmpz_equal};
+use flint_sys::fmpz::{fmpz, fmpz_cmp, fmpz_equal};
 use std::cmp::Ordering;
 
 impl PartialEq for Modulus {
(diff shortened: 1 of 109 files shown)


src/rational/q/cmp.rs
```diff
@@ -599,6 +671,53 @@ mod test_partial_ord {
     }
 }
 
+/// Test that the [`PartialOrd`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_ord_q_other {
+    use super::Q;
+    use crate::{integer::Z, integer_mod_q::Modulus};
+
+    // Ensure that the function can be called with several types
```
jnsiemer, line 680:
```suggestion
    /// Ensure that the function can be called with several types
```

src/macros/for_others.rs
```diff
@@ -96,6 +99,27 @@ macro_rules! implement_for_others {
...
+        $(#[doc(hidden)] impl PartialOrd<$source_type> for $type {
+            paste::paste! {
+                #[doc = "Documentation can be found at [`" $type "::partialEq`]."]
+                fn partial_cmp(&self, other: &$source_type) -> Option<Ordering> {
+                    self.partial_cmp(&$bridge_type::from(other))
+                }
+            }
+        }
+
+        #[doc(hidden)] impl PartialOrd<$type> for $source_type {
```
jnsiemer, line 114: Can't we take two additional input parameters for the trait and the function s.t. we don't have to write another part to this macro?

src/integer_mod_q/modulus/cmp.rs
```diff
@@ -196,6 +324,46 @@ mod test_partial_ord {
     }
 }
 
+/// Test that the [`PartialOrd`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_ord_modulus_other {
+    use super::Modulus;
+    use crate::integer::Z;
+
+    // Ensure that the function can be called with several types
```
jnsiemer, line 333:
```suggestion
    /// Ensure that the function can be called with several types
```

src/integer_mod_q/modulus/cmp.rs
```diff
@@ -159,6 +235,58 @@ mod test_eq {
...
+        assert!(2i64 == modulus);
+        assert!(2u64 == modulus);
+
+        assert!(&modulus == &z);
+        assert!(&z == &modulus);
+        assert!(&modulus == &2i8);
+        assert!(&2i8 == &modulus);
+    }
+
+    // Ensure that large values are compared correctly
```
jnsiemer, line 279:
```suggestion
    /// Ensure that large values are compared correctly
```

src/integer_mod_q/modulus/cmp.rs
```diff
@@ -159,6 +235,58 @@ mod test_eq {
     }
 }
 
+/// Test that the [`PartialEq`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_eq_modulus_other {
+    use super::Z;
+    use crate::integer_mod_q::Modulus;
+
+    // Ensure that the function can be called with several types
```
jnsiemer, line 244:
```suggestion
    /// Ensure that the function can be called with several types
```

jnsiemer (approved):
Looks good!

src/macros/for_others.rs
Marcel583: It is not so simple, because for example the mul trait needs an output type definition and PartialEq and PartialOrd have different output types.

PR #461, opened 2025-02-25
Make `DiscreteGaussianIntegerSampler` publicly available

jnsiemer:
**Description**
This PR implements...
- [x] makes DGIS publicly available.

This is required to enable quicker preimage-sampling for G-trapdoors in `qfall-crypto`.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -16,6 +16,6 @@ pub(crate) mod dimensions;
 mod factorization;
 pub mod index;
 pub(crate) mod parse;
-pub(crate) mod sample;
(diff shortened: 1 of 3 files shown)


Marvin-Beckmann (approved):
I am not sure, if we should make all parameters of the sampler public, as changing them would cause inconsistencies with the hash table.
I see two options:
1. Either make the variables of the struct non public to prevent them from being changed (and add getters if you really need to get access to them)
2. Make very huge declaimers in the documentation for that case.

I am fine either way I think, but I would definitely include some comment on it

PR #462, opened 2025-02-25
NTT Implementation for PolynomialRingZq

Marvin-Beckmann:
**Description**

This PR implements the NTT for
- [x] Cyclic Modulus
- [x] Negacyclic Modulus
- [x] multiplication of PolynomialRingZq

Even though I included several optimisations I found, I did not manage to make the implication faster than direct multiplication using FLINT, where the multiplication is already very fast.
I included benchmarks and added an additional disclaimer in the comment of the NTT-multiplication that is is not slower.
I tested it for common parameter-sets that I have found.
I found that the overhead necessary for multithreading also exceeds the runtime of the multiplication, so this was not an option to reduce the runtime, but for future potential optimisations I kept the framework to enable multi-threading within the code-base.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/ntt_basis_polynomial_ring_zq/intt.rs
+++ b/src/integer_mod_q/ntt_basis_polynomial_ring_zq/intt.rs
@@ -232,6 +232,17 @@ mod test_intt {
         ConvolutionType, Modulus, NTTBasisPolynomialRingZq, PolyOverZq, Zq,
     };
 
+    /// Tests a loop of NTT and INTT application.
+    #[test]
+    fn example_loop_ntt_intt() {
+        let poly = PolyOverZq::from_str("4  2532 2542 653 8 mod 7681").unwrap();
+        let modulus = Modulus::from(7681);
+
+        let ntt_basis = NTTBasisPolynomialRingZq::init(4, 3383, &modulus, ConvolutionType::Cyclic);
+
+        assert_eq!(poly, ntt_basis.intt(ntt_basis.ntt(&poly)));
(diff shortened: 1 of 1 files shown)


src/integer_mod_q/ntt_basis_polynomial_ring_zq/intt.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        out
+    }
+}
+
+/// This function essentially computes the included butterliy computations for each provided
+/// chunk.
+/// The chunk is double the size of the stride.
+/// The computation currently performs the standard butterly operation from Gentleman-Sande.
```
LPorzenheim, line 136: butterfly, see first sentence as well

src/integer_mod_q/ntt_basis_polynomial_ring_zq/intt.rs
```diff
@@ -0,0 +1,268 @@
...
+                &mut x.value.value,
+                &x.value.value,
+                &n_inv.value.value,
+                n_inv.modulus.get_fmpz_mod_ctx_struct(),
+            )
+        }
+    }
+    res
+}
+
```
LPorzenheim, line 225: Tests missing for intt working correctly as well as different modulus (as in ntt)

Marvin-Beckmann: added two more tests

LPorzenheim (approved):

PR #463, opened 2025-02-26
Multi-Threading for Sample_z

Marvin-Beckmann:
…reading

**Description**

This PR implements
- [x] a method to utilise multi-threading for sample_z, and use it for samples for matrices and polynomials.

Improvements, measured on MAC
```
SampleZ narrow 10,000   time:   [1.7816 ms 1.7972 ms 1.8158 ms]
                        change: [-83.647% -83.539% -83.425%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild

SampleZ wide single     time:   [8.5441 µs 8.5694 µs 8.5949 µs]
                        change: [-0.4292% +0.1489% +0.7009%] (p = 0.61 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) high mild
  4 (4.00%) high severe

SampleZ narrow single   time:   [4.2505 µs 4.2601 µs 4.2714 µs]
                        change: [-2.2812% -1.8075% -1.3209%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  4 (4.00%) high mild
  3 (3.00%) high severe
```
- [ ] let benchmark run on other devices
- [ ] consider speed for small matrices

**Testing**

**Checklist:**

### DIFF
--- a/src/utils/sample/discrete_gauss.rs
+++ b/src/utils/sample/discrete_gauss.rs
@@ -24,6 +24,10 @@ use crate::{
     traits::{GetNumColumns, GetNumRows, Pow},
 };
 use rand::RngCore;
+use rayon::{
+    current_num_threads,
+    iter::{IntoParallelIterator, ParallelIterator},
+};
 use serde::Serialize;
 use std::collections::HashMap;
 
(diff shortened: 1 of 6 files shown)


src/utils/sample/discrete_gauss.rs
```diff
@@ -172,6 +176,65 @@ impl DiscreteGaussianIntegerSampler {
...
+        if nr_threads == 1 || nr_samples < 10 {
+            // no multithreading
+            (0..nr_samples).map(|_| self.sample_z()).collect()
+        } else {
+            // with multithreading
+            let entries_per_thread = nr_samples / nr_threads;
+            let remainder = nr_samples % nr_threads;
+            (0..nr_threads)
+                .into_par_iter()
+                .map(|thread_i| {
```
jnsiemer, line 221: Just to provide another idea to make things quicker (although this might not be `rayon`'s way of parallelising things).
Currently, you split the workload into similar-sized buckets - implicitely making the assumption that each bucket will roughly take the same duration on each thread. This assumption should be correct in this case for larger bucket sizes, but for smaller sizes, it might be quicker to just submit tasks to a pool of threads, where each thread collects a new task once it has finished the current one.

src/utils/sample/discrete_gauss.rs
```diff
@@ -172,6 +176,65 @@ impl DiscreteGaussianIntegerSampler {
...
+                        entries_per_thread
+                    };
+                    (0..entries_thread_i)
+                        .map(|_| dgis_thread.sample_z())
+                        .collect()
+                })
+                .reduce(Vec::new, |mut a, mut b| {
+                    a.append(&mut b);
+                    a
+                })
```
jnsiemer, line 235: I haven't looked at the flamegraph, but considering my current experience with our library, setting the values of the matrix has a significant overhead. Furthermore, joining and iterating vectors shouldn't be the fastest thing in the world.
Could it be possible that it would be quicker to set the entry in the matrix directly after sampling it - sharing the matrix in an `Arc`?

src/utils/sample/discrete_gauss.rs
Marvin-Beckmann: I looked again, but I was not able to find a better solution for it, and given that the function will probably not be called with small numbers of samples - I think the current implementation is reasonable.
The problem with the dynamic approach is that I was not able to find a good way to also distribute the integer sampler, and additionally, this also provides an overhead with more threadmanagement, which might also increase the runtime due to the dynamic distribution of tasks.

src/utils/sample/discrete_gauss.rs
Marvin-Beckmann: I will check later, but the join/reduce is probably efficient, as it moves the values.
I would assume that the additional management cost of Arc will exceed the runtime of the collect.

PR #465, opened 2025-02-26
`MatQ` Multiplication with f64 precision

jnsiemer:
**Description**

This PR implements...
- [x] `mul_f64_unchecked` 
- [x] `collect_entries_f64`
- [x] a benchmark for MatQ multiplication

for `MatQ`.

For matrices with larger nominators & denominators, the MatQ::mul that doesn't allow for small precision loss is sometimes too inefficient. `mul_f64_unchecked` changes this by allowing for some loss of precision. This is a naive implementation of matrix multiplication with f64. It performs in the provided benchmark roughly 95% better than MatQ::mul.

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/mat_q/arithmetic/mul.rs
+++ b/src/rational/mat_q/arithmetic/mul.rs
@@ -14,7 +14,7 @@ use crate::integer::MatZ;
 use crate::macros::arithmetics::{
     arithmetic_trait_borrowed_to_owned, arithmetic_trait_mixed_borrowed_owned,
 };
-use crate::traits::{GetNumColumns, GetNumRows};
+use crate::traits::{GetNumColumns, GetNumRows, SetEntry};
 use flint_sys::fmpq_mat::{fmpq_mat_mul, fmpq_mat_mul_fmpz_mat};
 use std::ops::Mul;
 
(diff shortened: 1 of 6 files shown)


src/rational/mat_q/get.rs
```diff
@@ -289,6 +289,40 @@ impl MatQ {
...
+    /// use qfall_math::rational::MatQ;
+    /// use std::str::FromStr;
+    ///
+    /// let mat = MatQ::from_str("[[1/1, 2],[3/1, 4],[5/1, 6]]").unwrap();
+    ///
+    /// let entries_f64 = mat.collect_entries_f64();
+    ///
+    /// assert_eq!(entries_f64[0][1], 2.0);
+    /// ```
+    pub fn collect_entries_f64(&self) -> Vec<Vec<f64>> {
```
Marvin-Beckmann, line 308: what happens if the entries are too large to fit into an f64?

src/rational/mat_q/get.rs
```diff
@@ -684,4 +718,28 @@ mod test_collect_entries {
         assert_eq!(entries_2[1].num.0, -1);
         assert_eq!(entries_2[1].den.0, 2);
     }
+
+    #[test]
+    fn all_entries_collected_f64() {
```
Marvin-Beckmann, line 723: Large numbers are missing for behavioral checks

src/rational/mat_q/arithmetic/mul.rs
```diff
@@ -94,6 +94,53 @@ impl Mul<&MatZ> for &MatQ {
...
+    /// ```
+    /// use qfall_math::integer::MatZ;
+    /// let mat = MatZ::sample_uniform(3, 3, -256, 256).unwrap().inverse().unwrap();
+    ///
+    /// let mat_inv_sqrd = mat.mul_f64_unchecked(&mat);
+    /// ```
+    ///
+    /// # Panics ...
+    /// - if the dimensions of `self` and `other` do not match for multiplication.
+    pub fn mul_f64_unchecked(&self, other: &Self) -> MatQ {
```
Marvin-Beckmann, line 122: what happens if there is an overflow in the multiplication or addition in f64?
You mentioned unchecked in terms of dimension, but this also does not work for very large values

Marvin-Beckmann (changes requested):
Not all cases in the function description are covered where errors can occur

src/rational/mat_q/arithmetic/mul.rs
jnsiemer: I've added some comments. The deviation is worse than I thought for large values (size 2^64 can deviate by roughly 1000). I've addressed it in the doc-comment explicitely as a warning (and split them up s.t. they should be more recognizable).
I've thought about making this function `unsafe` to make the user read the doc-comment before using it. However, currently we're using unsafe functions to mark behaviour that might introduce memory leaks and when someone uses a function with f64 explicitely stated in it, he/she should be able to figure out that there can be some deviation. Would you make it unsafe or keep it the way it is?

Marvin-Beckmann: I would keep it as it is

src/rational/mat_q/arithmetic/mul.rs
```diff
@@ -106,7 +106,14 @@ impl MatQ {
     /// or entries with small numerators and denominators. This function becomes more
     /// efficient once `self` or `other` has entries with large numerators and denominators
     /// as FLINT's implementation does not allow any loss of precision.
-    /// Furthermore, this function is unchecked, i.e. the user is expected to align matrix
+    ///
+    /// **WARNING:** Please be aware that the deviation of the representation of the matrices' entries as [`f64`]
+    /// will scale with the size of the entries, e.g. an entry within the size of `2^{64}`
+    /// might deviate from the original value by a distance of `1_000`. This loss of precision
+    /// might be aggravated by this matrix multiplication with a factor of `self.get_num_columns()`
+    /// for each entry in the resulting matrix.
```
Marvin-Beckmann, line 114: even if you multiply two matrices with entries in 2**32, you can easily have overflow here.
So maybe make this somewhat more precise

Marvin-Beckmann (approved):

src/rational/mat_q/arithmetic/mul.rs
jnsiemer: I've addressed this in the `Panics` section with
```
/// if any result during the naive computation of matrix multiplication is larger than [`f64::MAX`] or smaller than [`f64::MIN`].
```
However, you have to keep in mind that [f64::MAX] is 1.7976931348623157 × 10^308, which is huge. It is way larger than any u64::MAX value. This is due to the fact that it doesn't guarantee precision on an integer level a some point. Hence, I think we should be fine with addressing this very odd case in the Panic section.

PR #478, opened 2025-03-16
`AddAssign` for Matrices

jnsiemer:
**Description**
This PR implements `AddAssign` for ...
- [x] MatZ
- [x] MatZq
- [x] MatQ
- [x] MatPolyOverZ
- [x] MatPolynomialRingZq

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
+++ b/src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
@@ -15,9 +15,7 @@ use crate::{
         arithmetic_assign_trait_borrowed_to_owned, arithmetic_trait_borrowed_to_owned,
         arithmetic_trait_mixed_borrowed_owned,
     },
-    traits::MatrixDimensions,
 };
-use flint_sys::fmpz_poly_mat::fmpz_poly_mat_add;
 use std::ops::{Add, AddAssign};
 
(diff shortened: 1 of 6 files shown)


src/integer_mod_q/mat_polynomial_ring_zq/arithmetic/add.rs
```diff
@@ -12,10 +12,75 @@ use super::super::MatPolynomialRingZq;
...
+            );
+        }
+
+        unsafe {
+            fmpz_poly_mat_add(
+                &mut self.matrix.matrix,
+                &self.matrix.matrix,
+                &other.matrix.matrix,
+            )
+        };
```
Marcel583, line 73: Maybe use add_assign for MatPolyOverZ directly (together with the modulus check)

src/rational/mat_q/arithmetic/add.rs
```diff
@@ -10,12 +10,76 @@
...
+            panic!(
+                "Tried to add a '{}x{}' matrix and a '{}x{}' matrix.",
+                self.get_num_rows(),
+                self.get_num_columns(),
+                other.get_num_rows(),
+                other.get_num_columns()
+            );
+        }
+
+        // there should be a better way than this, but I haven't found it
```
Marcel583, line 75: I don't know if we want this comment here. Maybe put it in the Backlog

PR #480, opened 2025-03-17
PartialEq between MatZ and MatQ

Marcel583:
**Description**

This PR implements partialEq between MatZ and MatQ.

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/mat_q/cmp.rs
+++ b/src/rational/mat_q/cmp.rs
@@ -1,4 +1,4 @@
-// Copyright © 2023 Sven Moog
+// Copyright © 2023 Sven Moog, Marcel Luca Schmidt
 //
 // This file is part of qFALL-math.
 //
@@ -9,6 +9,11 @@
 //! This module contains implementations for comparison of [`MatQ`].
 
 use super::MatQ;
+use crate::{
+    integer::MatZ,
+    macros::for_others::implement_trait_reverse,
+    traits::{GetEntry, GetNumColumns, GetNumRows},
+};
 use flint_sys::fmpq_mat::fmpq_mat_equal;
 
 impl PartialEq for MatQ {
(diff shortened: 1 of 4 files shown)


src/rational/mat_q/cmp.rs
```diff
@@ -49,6 +54,63 @@ impl PartialEq for MatQ {
...
+        let num_rows = self.get_num_rows();
+        let num_cols = self.get_num_columns();
+
+        if num_rows != other.get_num_rows() || num_cols != other.get_num_columns() {
+            return false;
+        }
+
+        for i in 0..num_rows {
+            for j in 0..num_cols {
+                if self.get_entry(i, j).unwrap() != other.get_entry(i, j).unwrap() {
```
jnsiemer, line 102:
```suggestion
                if unsafe { self.get_entry_unchecked(i, j) != other.get_entry_unchecked(i, j) } {
```

src/rational/mat_q/cmp.rs
```diff
@@ -9,6 +9,11 @@
 //! This module contains implementations for comparison of [`MatQ`].
 
 use super::MatQ;
+use crate::{
+    integer::MatZ,
+    macros::for_others::implement_trait_reverse,
+    traits::{GetEntry, GetNumColumns, GetNumRows},
```
jnsiemer, line 15: You'll need to rebase before merging - then these traits will be named differently

src/rational/mat_q/cmp.rs
```diff
@@ -49,6 +54,63 @@ impl PartialEq for MatQ {
...
+            return false;
+        }
+
+        for i in 0..num_rows {
+            for j in 0..num_cols {
+                if self.get_entry(i, j).unwrap() != other.get_entry(i, j).unwrap() {
+                    return false;
+                }
+            }
+        }
```
jnsiemer, line 106: I might have a quicker solution for this.

```
let mut other_matq = MatQ::new(other.get_num_rows(), other.get_num_columns());
unsafe { fmpq_mat_set_fmpz_mat_div_fmpz(&mut other_matq.matrix, &other.matrix, &fmpz(1)) };
1 != unsafe { fmpq_mat_equal(&other_matq.matrix, self.matrix) }
```

Might we worth a check how quick your solution is compared to this one. Theoretically, your's should be quicker, but every call to `get_entry` has some overhead that `fmpq_mat_equal` and `fmpq_mat_set_fmpz_mat_div_fmpz` probably don't have.

src/rational/mat_q/cmp.rs
```diff
@@ -116,3 +178,33 @@ mod test_partial_eq {
...
+
+        assert!(q == z);
+        assert!(z == q);
+        assert!(&q == &z);
+        assert!(&z == &q);
+    }
+
+    // Ensure that large values are compared correctly
+    #[test]
+    fn equal_large() {
```
jnsiemer, line 204:
```suggestion
    /// Ensure that large values are compared correctly.
    #[test]
    fn equal_large() {
```
What about inequalitites? A quick test for that would be nice.

src/rational/mat_q/cmp.rs
```diff
@@ -116,3 +178,33 @@ mod test_partial_eq {
...
 }
+
+/// Test that the [`PartialEq`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_eq_q_other {
+    use super::MatQ;
+    use crate::integer::MatZ;
+    use std::str::FromStr;
+
+    // Ensure that the function can be called with several types
```
jnsiemer, line 189:
```suggestion
    /// Ensure that the function can be called with several types.
```

jnsiemer (approved):
Looks good except for a few minor issues - and please rebase before merging! Otherwise, the next `dev` pipeline run will fail.

src/rational/mat_q/cmp.rs
Marcel583: The current one is slightly faster, so I would keep it

PR #481, opened 2025-03-18
PartialEq between PolyOverZ and PolyOverQ

Marcel583:
**Description**

This PR implements PartialEq between PolyOverZ and PolyOverQ.

**Testing**

**Checklist:**

### DIFF
--- a/src/rational/poly_over_q/cmp.rs
+++ b/src/rational/poly_over_q/cmp.rs
@@ -285,7 +285,7 @@ mod test_partial_eq_q_other {
     use crate::integer::PolyOverZ;
     use std::str::FromStr;
 
-    // Ensure that the function can be called with several types
+    /// Ensure that the function can be called with several types.
     #[test]
     #[allow(clippy::op_ref)]
     fn availability() {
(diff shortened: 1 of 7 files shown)


src/rational/poly_over_q/cmp.rs
```diff
@@ -221,3 +277,33 @@ mod test_partial_eq {
...
 }
+
+/// Test that the [`PartialEq`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_eq_q_other {
+    use super::PolyOverQ;
+    use crate::integer::PolyOverZ;
+    use std::str::FromStr;
+
+    // Ensure that the function can be called with several types
```
jnsiemer, line 288:
```suggestion
    /// Ensure that the function can be called with several types.
```

src/rational/poly_over_q/cmp.rs
```diff
@@ -221,3 +277,33 @@ mod test_partial_eq {
...
+        let q = PolyOverQ::from_str("4  1 2 3 4").unwrap();
+        let z = PolyOverZ::from_str("4  1 2 3 4").unwrap();
+
+        assert!(q == z);
+        assert!(z == q);
+        assert!(&q == &z);
+        assert!(&z == &q);
+    }
+
+    // Ensure that large values are compared correctly
```
jnsiemer, line 301:
```suggestion
    /// Ensure that large values are compared correctly.
```

src/rational/poly_over_q/cmp.rs
```diff
@@ -221,3 +277,33 @@ mod test_partial_eq {
         assert_ne!(one, two);
     }
 }
+
+/// Test that the [`PartialEq`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_eq_q_other {
```
jnsiemer, line 283: Please add a test for a polynomial that isn't equal to the other one (same number of coefficients).
Furthermore, please add two tests for different numbers of coefficients (one that is equal with 0 as largest coefficient, one for different ones but different lengths). - To make sure that the function works for polynomials of different degrees, too.

jnsiemer (approved):
Looks good. Please add the tests before merging.

PR #482, opened 2025-03-18
PartialEq between PolyOverZq and ModPoly

Marcel583:
**Description**

This PR implements PartialEq between PolyOverZq and ModPoly.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/cmp.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/cmp.rs
@@ -10,6 +10,10 @@
 //! This uses the traits from [`std::cmp`].
 
 use super::ModulusPolynomialRingZq;
+use crate::{
+    integer::Z, integer_mod_q::PolyOverZq, macros::for_others::implement_trait_reverse,
+    traits::GetCoefficient,
+};
 use flint_sys::{fmpz::fmpz_equal, fmpz_mod_poly::fmpz_mod_poly_equal};
 
 impl PartialEq for ModulusPolynomialRingZq {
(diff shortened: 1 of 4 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/cmp.rs
```diff
@@ -234,3 +293,33 @@ mod test_partial_eq {
...
+        let q = ModulusPolynomialRingZq::from_str("4  1 2 3 4 mod 17").unwrap();
+        let z = PolyOverZq::from_str("4  1 2 3 4 mod 17").unwrap();
+
+        assert!(q == z);
+        assert!(z == q);
+        assert!(&q == &z);
+        assert!(&z == &q);
+    }
+
+    // Ensure that large values are compared correctly
```
jnsiemer, line 316:
```suggestion
    /// Ensure that large values are compared correctly.
```

src/integer_mod_q/modulus_polynomial_ring_zq/cmp.rs
```diff
@@ -234,3 +293,33 @@ mod test_partial_eq {
         assert_ne!(first, second);
     }
 }
+
+/// Test that the [`PartialEq`] trait is correctly implemented.
+#[cfg(test)]
+mod test_partial_eq_q_other {
```
jnsiemer, line 299: Please add a test for a polynomial that isn't equal to the other one (same number of coefficients).
Furthermore, please add two tests for different numbers of coefficients (one that is equal with 0 as largest coefficient, one for different ones but different lengths). - To make sure that the function works for polynomials of different degrees, too.

jnsiemer (approved):
Looks good. Please add the tests before merging.

PR #485, opened 2025-03-19
Debug trait with meaningful output

Marcel583:
**Description**

This PR implements the Debug trait for all types by using our to_string method, so users do not get pointers as a debug message.

- [x] matrices are only printed in a limited representation
- [x] the previously present storage information remains
- [x] doc comment for matrix debug print method

**Testing**

**Checklist:**


src/rational/poly_over_q/to_string.rs
```diff
@@ -95,9 +95,6 @@ impl PolyOverQ {
     ///
     /// let decimal_repr = poly.to_string_decimal(3);
     /// ```
-    ///
-    /// # Panics ...
-    /// - if any entry of the polynomial can't be represented as a [`f64`].
```
jnsiemer, line 100: It probably needs a slightly more careful wording, but this function can panic if the value is simply too large to be represented by a [`f64`]. See [f64::MAX] - although if someone reaches this value, something's gone wrong anyway :)

src/rational/mat_q/to_string.rs
```diff
@@ -97,9 +98,6 @@ impl MatQ {
     ///
     /// let decimal_repr = matrix.to_string_decimal(3);
     /// ```
-    ///
-    /// # Panics ...
-    /// - if any entry of the matrix can't be represented as an [`f64`].
```
jnsiemer, line 102: ^^

src/rational/q.rs
```diff
@@ -71,7 +72,18 @@ mod unsafe_functions;
...
-#[derive(Debug)]
 pub struct Q {
     pub(crate) value: fmpq,
 }
+
+impl fmt::Debug for Q {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Q {{value_f64(may be rounded): {}, value: {}, storage: {{value: {:?}}}}}",
```
jnsiemer, line 83: Why don't you use the `.to_string_decimal` function here with a fixed precision?

jnsiemer (approved):
Some minor things to do / to think about

src/rational/poly_over_q/to_string.rs
Marvin-Beckmann: where exactly does the panic occur? I cant see it anywhere? In the form method for f64 there does not seem to be any panic

src/rational/q.rs
Marvin-Beckmann: fixed; I did it this way because there is only one value to represent, and I was not concerned by the number of elements that needed printing.
However, due to consistency I changed it to 5 decimals now

src/rational/poly_over_q/to_string.rs
jnsiemer: Ok, I've just tested it. It doesn't result in a panic.
```
#[test]
fn a() {
    let a = f64::MAX;
    let rational_0 = Q::from(a);
    let rational_1 = rational_0 * 15;
    let b = f64::from(&rational_1);
    println!("{b}");
}
```
But `b` prints `inf` in this case, i.e. the value is so large that it's not representable by a `f64`. This behaviour seems reasonable to me and doesn't need additional comments.

PR #486, opened 2025-03-20
Scalar multiplication

Marcel583:
**Description**

This PR implements scalar multiplication for all types.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/poly_over_z/arithmetic/mul_scalar.rs
+++ b/src/integer/poly_over_z/arithmetic/mul_scalar.rs
@@ -1,0 +1,271 @@
+// Copyright © 2025 Marcel Luca Schmidt
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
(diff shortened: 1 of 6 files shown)


src/integer/poly_over_z/arithmetic/mul_scalar.rs
```diff
@@ -0,0 +1,277 @@
...
+    ///
+    /// let poly_2 = &poly_1 * &rational;
+    /// ```
+    fn mul(self, scalar: &Q) -> PolyOverQ {
+        let mut out = PolyOverQ::default();
+        unsafe {
+            fmpq_poly_scalar_mul_fmpq(&mut out.poly, &PolyOverQ::from(self).poly, &scalar.value);
+        }
+        out
+    }
```
Marvin-Beckmann, line 136: Make PolyQ as you have done, and then just return the multiplication of it with the scalar value to reduce access to the FFI.

src/rational/poly_over_q/arithmetic/div_scalar.rs
Marvin-Beckmann, line 1: what happens if we divide by 0? just out of curiosity

Marvin-Beckmann (approved):
Thanks for doing this tedious work. Looks good.

PR #491, opened 2025-03-22
Add `unsafe` Setters

jnsiemer:
**Description**
This PR implements...
- [x] public unsafe set_<fmpz_struct_name> for every underlying fmpz struct.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/z/unsafe_functions.rs
+++ b/src/integer/z/unsafe_functions.rs
@@ -22,7 +22,7 @@ unsafe_setter!(Z, value, fmpz, fmpz_clear);
 mod test_get_fmpz {
     use super::Z;
 
-    /// Checks availability of the setter for [`Z::value`]
+    /// Checks availability of the getter for [`Z::value`]
     /// and its ability to be modified.
     #[test]
     #[allow(unused_mut)]
--- a/src/integer/z/unsafe_functions.rs
+++ b/src/integer/z/unsafe_functions.rs
@@ -9,16 +9,20 @@
 //! This module contains public functions that enable access to underlying
 //! [FLINT](https://flintlib.org/) structs. Therefore, they require to be unsafe.
 
-use crate::{integer::Z, macros::unsafe_passthrough::unsafe_getter};
-use flint_sys::fmpz::fmpz;
+use crate::{
+    integer::Z,
+    macros::unsafe_passthrough::{unsafe_getter, unsafe_setter},
+};
+use flint_sys::fmpz::{fmpz, fmpz_clear};
 
 unsafe_getter!(Z, value, fmpz);
+unsafe_setter!(Z, value, fmpz, fmpz_clear);
 
 #[cfg(test)]
 mod test_get_fmpz {
     use super::Z;
 
-    /// Checks availability of the getter for [`Z::value`]
+    /// Checks availability of the setter for [`Z::value`]
     /// and its ability to be modified.
     #[test]
     #[allow(unused_mut)]
@@ -28,6 +32,25 @@ mod test_get_fmpz {
         let mut fmpz_int = unsafe { integer.get_fmpz() };
 
         fmpz_int.0 = 2;
+
+        assert_eq!(Z::from(2), integer);
+    }
+}
+
+#[cfg(test)]
+mod test_set_fmpz {
+    use super::Z;
+    use flint_sys::fmpz::fmpz;
+
+    /// Checks availability of the setter for [`Z::value`]
+    /// and its ability to modify [`Z`].
+    #[test]
+    #[allow(unused_mut)]
+    fn availability_and_modification() {
+        let mut integer = Z::from(1);
+        let b = fmpz(2);
+
+        unsafe { integer.set_fmpz(b) };
 
         assert_eq!(Z::from(2), integer);
     }
--- a/src/integer_mod_q/modulus/unsafe_functions.rs
+++ b/src/integer_mod_q/modulus/unsafe_functions.rs
@@ -11,9 +11,51 @@
 
 use super::Modulus;
 use crate::macros::unsafe_passthrough::unsafe_getter_mod;
-use flint_sys::fmpz_mod::fmpz_mod_ctx;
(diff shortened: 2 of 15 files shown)


src/integer/z/unsafe_functions.rs
```diff
@@ -9,16 +9,20 @@
...
 
 unsafe_getter!(Z, value, fmpz);
+unsafe_setter!(Z, value, fmpz, fmpz_clear);
 
 #[cfg(test)]
 mod test_get_fmpz {
     use super::Z;
 
-    /// Checks availability of the getter for [`Z::value`]
+    /// Checks availability of the setter for [`Z::value`]
```
Marcel583, line 25: But you check the getter here. So the initial comment should be correct

src/integer_mod_q/modulus/unsafe_functions.rs
```diff
@@ -33,3 +75,28 @@ mod test_get_fmpz_mod_ctx {
...
+        let mut flint_struct = unsafe {
+            fmpz_mod_ctx_init(flint_struct.as_mut_ptr(), &fmpz(2));
+            flint_struct.assume_init()
+        };
+
+        unsafe { modulus.set_fmpz_mod_ctx(flint_struct) };
+
+        assert_eq!(Modulus::from(2), modulus);
+    }
+}
```
Marcel583, line 102: How does it work for multiple references?
You say that the function only works if no other references exist/are used in the comment. But is there no way to rewrite the function such that it does? Or is it just too much for us to implement right now?
I tried this, but it did not work:

/// Checks if setter works for all references.

    #[test]
    #[allow(unused_mut)]
    fn counter_works() {
        let mut modulus_1 = Modulus::from(3);
        let mut modulus_2 = Modulus::from(3);
        let mut modulus_3 = modulus_1.clone();

        let mut flint_struct = MaybeUninit::uninit();
        let mut flint_struct = unsafe {
            fmpz_mod_ctx_init(flint_struct.as_mut_ptr(), &fmpz(2));
            flint_struct.assume_init()
        };

        unsafe { modulus_1.set_fmpz_mod_ctx(flint_struct) };

        assert_eq!(Modulus::from(2), modulus_1);
        assert_eq!(Modulus::from(3), modulus_2);
        assert_eq!(Modulus::from(2), modulus_3);
    }

src/integer_mod_q/modulus/unsafe_functions.rs
```diff
@@ -11,10 +11,52 @@
 
 use super::Modulus;
 use crate::macros::unsafe_passthrough::unsafe_getter_mod;
-use flint_sys::fmpz_mod::fmpz_mod_ctx;
+use flint_sys::fmpz_mod::{fmpz_mod_ctx, fmpz_mod_ctx_clear};
 
 unsafe_getter_mod!(Modulus, modulus, fmpz_mod_ctx);
 
+impl Modulus {
+    /// Returns a mutable reference to the field `modulus` of type [`Modulus`].
```
Marcel583, line 19: The function returns nothing. It sets the value
```suggestion
    /// Sets a mutable reference to the field `modulus` of type [`Modulus`] to a given `fmpz_mod_ctx`.
```

src/integer_mod_q/modulus/unsafe_functions.rs
```diff
@@ -11,10 +11,52 @@
...
+    ///
+    /// Parameters:
+    /// - `flint_struct`: value to set the attribute to
+    ///
+    /// **WARNING:** The returned struct is part of [`flint_sys`].
+    /// Any changes to this object are unsafe and may introduce memory leaks.
+    /// Please be aware that most moduli are shared across multiple instances and all
+    /// modifications of this struct will affect any other instance with a reference to this object.
```
Marcel583, line 27:
```suggestion
    /// **WARNING:** The set struct is part of [`flint_sys`].
    /// Any changes to this object are unsafe and may introduce memory leaks.
    /// Please be aware that most moduli are shared across multiple instances and all
    /// modifications of this struct will affect any other instance with a reference to this object.
```

src/integer_mod_q/modulus_polynomial_ring_zq/unsafe_functions.rs
```diff
@@ -11,10 +11,54 @@
 
 use super::ModulusPolynomialRingZq;
 use crate::macros::unsafe_passthrough::unsafe_getter_mod;
-use flint_sys::fq::fq_ctx_struct;
+use flint_sys::fq::{fq_ctx_clear, fq_ctx_struct};
 
 unsafe_getter_mod!(ModulusPolynomialRingZq, modulus, fq_ctx_struct);
 
+impl ModulusPolynomialRingZq {
+    /// Returns a mutable reference to the field `modulus` of type [`ModulusPolynomialRingZq`].
```
Marcel583, line 19:
```suggestion
    /// Sets a mutable reference to the field `modulus` of type [`ModulusPolynomialRingZq`] to a given `fq_ctx_struct`.
```

Marcel583 (commented):
Looks good. Just a question I have and some comment changes

src/integer_mod_q/modulus/unsafe_functions.rs
jnsiemer: I would have loved to implement this. Unfortunately, you can't change the object that a Reference Counter (Rc), which we implement our Modulus objects with, retrospectively.
As I wasn't able to find out which objects point to the Modulus object (I believe that the data-structure internally only keeps a reference from the object pointing to the Rc, which would be absolutely sensible), it could be possible that this is impossible if you don't want to do a search for a reference to that Rc and replace the pointer. Also in this case, I wouldn't know how to do it and it seems very inefficient.

Thus, I settled for this solution for now.

Marvin-Beckmann: Yes, not possible. You would need additional modifications to the baseline. The Rc provides immutable access to the content of the reference counter.
This is also why modification as here needs to drop the current RC and essentially needs to instantiate a new counter.

What you could do is provide a Rc to an fmpz_mod_ctx object instead of the context object itself - then you could get the desired behavior

Marvin-Beckmann (approved):

PR #497, opened 2025-04-13
Rework Getters for Matrices by accessing via Submatrices

Marvin-Beckmann:
**Description**

This PR reworks how we have implemented getters for rows and columns for matrices.
The new implementation is easier to maintain and has less code redundancy.
Additionally I...
- [x] added unchecked getters for: rows, columns and submatrices
- [x] went through our current codebase and changed the implementations to use the more efficient unchecked getters where it makes sense.

**Testing**

I did not add additional dedicated tests for the unchecked matrices, as I think they are sufficiently tested through the checked methods that call them

**Checklist:**

  - [x] The code provides good readability and maintainability s.t. it fulfils best practices like talking code, modularity, ...

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -95,8 +95,6 @@ where
     /// occur.
     unsafe fn get_entry_unchecked(&self, row: i64, column: i64) -> T;
 
-    // *** Automatically implemented functions
-
     /// Outputs a [`Vec<Vec<T>>`] containing all entries of the matrix s.t.
     /// any entry in row `i` and column `j` can be accessed via `entries[i][j]`
     /// if `entries = matrix.get_entries`.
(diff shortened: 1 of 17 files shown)


src/traits.rs
```diff
@@ -189,26 +192,72 @@ where
     /// Parameters:
     /// - `row`: specifies the row of the matrix to return
     ///
+    /// Negative indices can be used to index from the back, e.g., `-1` for
+    /// the last element.
+    ///
     /// Returns a row vector of the matrix at the position of the given
-    /// `row` or an error if specified row is not part of the matrix.
+    /// `row` or an error if the number of rows is
+    /// greater than the matrix or negative.
```
Marcel583, line 200: I would keep the original

src/traits.rs
```diff
@@ -240,7 +289,48 @@ where
...
+    /// occur.
+    unsafe fn get_submatrix_unchecked(
+        &self,
+        row_1: i64,
+        row_2: i64,
+        col_1: i64,
+        col_2: i64,
+    ) -> Self;
 
     // *** Automatically implemented functions
```
Marcel583, line 335: Delete this comment

Marcel583 (approved):

PR #498, opened 2025-04-14
Fix Overindentation marked by Clippy

Marvin-Beckmann:
The pipeline fails due to updated clippy-requirements.
The indentation now has a fixed format, and we were not following these guidelines.
I went through the code to fix them to make clippy happy again.


Marcel583 (approved):
Looks Great!

PR #499, opened 2025-04-14
Rework Base Error Handling

Marvin-Beckmann:
**Description**

To implement set_submatrix with the smallest possible overhead I have found a quite nice solution, but I need the ability to compare basetypes through a trait.
In that, I decided to model how we currently compare base errors (moduli errors) and made it its own trait.
We lose some precision in the comments, but the stacktrace also gives the information where the respective errors are from, and all the required debug information is still present.

This PR also improves code-maintainability and makes it easier to compare the base types going forward.

This PR implements...
- [x] the CompareBase trait for all matrices and additionally all  base types, where a base comparison makes sense, i.e. PolyZq, PolynomialRingZq, Zq

**Testing**

**Checklist:**


src/integer_mod_q/mat_polynomial_ring_zq/cmp.rs
```diff
@@ -0,0 +1,103 @@
...
+//! This module contains implementations for comparison of [`MatPolynomialRingZq`].
+
+use super::MatPolynomialRingZq;
+use crate::{error::MathError, traits::CompareBase};
+
+impl CompareBase for MatPolynomialRingZq {
+    /// Compares the moduli of the two elements.
+    ///
+    /// Parameters:
+    /// - `other`: The other objects whose base is compared to `self`
```
Marcel583, line 18:
```suggestion
    /// - `other`: The other object whose base is compared to `self`
```

Marcel583 (approved):

PR #500, opened 2025-04-14
Implement SetSubmatrix Trait for all Matrices

Marvin-Beckmann:
**Description**

This PR implements...
- [x] the set submatrix trait for all matrices (test + documentation still missing)
- [x] uses generalized generic implemenetations for set_row and set_column to increase maintainability
- [x] optimizes internal implementation by using unchecked setter

**Testing**

- [x] I added basic working examples (possibly in doc-comment)
- [x] I added tests for large (pointer representation) values
- [x] I triggered all possible errors in my test in every possible way
- [x] I included tests for all reasonable edge cases

**Checklist:**

- [x] I have performed a self-review of my own code
  - [x] The code provides good readability and maintainability s.t. it fulfills best practices like talking code, modularity, ...
    - [x] The chosen implementation is not more complex than it has to be
  - [x] My code should work as intended and no side effects occur (e.g. memory leaks)
  - [x] The doc comments fit our style guide
  - [x] I have credited related sources if needed

PR #502, opened 2025-04-20
Rework Coefficient Polynomials

Marvin-Beckmann:
**Description**

Similar to the previous PR, this PR reduces some code base for easier maintainability and adds unchecked methods.

- [x] unchecked getter and setter for polynomials
- [x] unchecked getter and setter for entries of matrices
- [x] a macro for compare base for easier implementations, such that it can be used for the polynomials + tests for availability
- [x] added a missing set_coefficient for PolynomialRingZq
- [x] use unchecked getter and setters in our code to improve the efficiency of our internal implementations.

For now I decided to only add the compare base types that were meaningful in my opinion, i.e. where we need it in our internal implementations.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/poly_over_z/coefficient_embedding.rs
+++ b/src/integer/poly_over_z/coefficient_embedding.rs
@@ -100,8 +100,7 @@ impl FromCoefficientEmbedding<&MatZ> for PolyOverZ {
         );
         let mut out = PolyOverZ::default();
         for i in 0..embedding.get_num_rows() {
-            out.set_coeff(i, unsafe { embedding.get_entry_unchecked(i, 0) })
-                .unwrap()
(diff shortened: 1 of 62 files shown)


src/integer/poly_over_z/coefficient_embedding.rs
```diff
@@ -57,7 +57,7 @@ impl IntoCoefficientEmbedding<MatZ> for &PolyOverZ {
         );
         let mut out = MatZ::new(size, 1);
         for j in 0..size {
-            let coeff = self.get_coeff(j).unwrap();
+            let coeff = unsafe { self.get_coeff_unchecked(j) };
```
3pmile, line 60: Same question as in https://github.com/qfall/math/pull/500, why do we want unsafe versions?

Traits could panic instead of Result. This would reduce the need of unwraps.

3pmile (approved):

PR #528, opened 2026-04-27
Add Hamming weight

jnsiemer:
**Description**
This PR implements...
- [x] a function to determine the hamming weight for all polynomials and matrix types.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer/mat_poly_over_z/norm.rs
+++ b/src/integer/mat_poly_over_z/norm.rs
@@ -90,8 +90,8 @@ impl MatPolyOverZ {
         max_norm
     }
 
-    /// Outputs the hamming weight of `self`, i.e. it computes the sum of all non-zero
-    /// coefficients in each polynomial in the matrix.
+    /// Outputs the hamming weight of `self`, i.e. it computes the number of non-zero
+    /// coefficients across all polynomials in the matrix.
     ///
     /// # Examples
     /// ```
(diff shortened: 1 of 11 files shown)


src/integer/mat_poly_over_z/norm.rs
```diff
@@ -89,6 +89,31 @@ impl MatPolyOverZ {
         }
         max_norm
     }
+
+    /// Outputs the hamming weight of `self`, i.e. it computes the sum of all non-zero
+    /// coefficients in each polynomial in the matrix.
```
Marvin-Beckmann: normally, the hamming weight is only defined as the number of non-zero coefficients, and not the sum of the non-zero coefficients. Typically the HW is also used in cases of binary vectors, so that could be a reason. Anyways, I think this definition as is, is confusing in regards to usage

src/integer_mod_q/mat_zq/norm.rs
```diff
@@ -89,6 +90,33 @@ impl MatZq {
         }
         max_norm
     }
+
+    /// Outputs the hamming weight of `self`, i.e. it returns the number of
+    /// non-zero entries in the matrix.
```
Marvin-Beckmann, line 95: why is the definition of hamming weight different here?

Marvin-Beckmann (changes requested):
please check for consistency. If this intended, then some renaming should be done in my opinion

src/integer/mat_poly_over_z/norm.rs
jnsiemer: This definition is wrong. I meant to say "the sum of number of non-zero coefficients..." to be explicit. However, the updated doc comment is precise.

Marvin-Beckmann (approved):

PR #532, opened 2026-05-06
Add `change_modulus` and `change_q`

jnsiemer:
**Description**

This PR implements...
- [x] `change_modulus`
- [x] `change_q`

for all non-NTT Zq datatypes.

**Testing**

**Checklist:**

### DIFF
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
@@ -8,7 +8,10 @@
 
 //! Implementations to manipulate a [`ModulusPolynomialRingZq`].
 
-use crate::integer_mod_q::{Modulus, ModulusPolynomialRingZq, PolyOverZq, modulus_polynomial_ring_zq::from::non_zero_positions};
+use crate::integer_mod_q::{
+    Modulus, ModulusPolynomialRingZq, PolyOverZq,
+    modulus_polynomial_ring_zq::from::non_zero_positions,
+};
 use std::rc::Rc;
 
 impl ModulusPolynomialRingZq {
@@ -40,7 +43,7 @@ impl ModulusPolynomialRingZq {
         let new_poly_zq =
             PolyOverZq::from((self.get_representative_least_nonnegative_residue(), modulus));
         self.non_zero = non_zero_positions(&new_poly_zq);
-        
+
         self.modulus = Rc::new(new_poly_zq);
     }
 }
@@ -96,7 +99,7 @@ mod test_change_q {
 
         modulus.change_q(&q);
 
-        assert!(modulus.non_zero.is_empty());
+        assert_eq!(vec![0, 2, 4, 5], modulus.non_zero);
         assert!(modulus.ntt_basis.is_none());
     }
 }
--- a/src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
+++ b/src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
@@ -1,0 +1,101 @@
+// Copyright 2026 Jan Niklas Siemer
+//
+// This file is part of qFALL-math.
+//
+// qFALL-math is free software: you can redistribute it and/or modify it under
+// the terms of the Mozilla Public License Version 2.0 as published by the
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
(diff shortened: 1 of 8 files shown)


src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,101 @@
...
+// Mozilla Foundation. See <https://mozilla.org/en-US/MPL/2.0/>.
+
+//! Implementations to manipulate a [`ModulusPolynomialRingZq`].
+
+use crate::integer_mod_q::{Modulus, ModulusPolynomialRingZq, PolyOverZq};
+use std::rc::Rc;
+
+impl ModulusPolynomialRingZq {
+    /// Changes the modulus `q` of the given [`ModulusPolynomialRingZq`] to the new modulus `q`
+    /// and resets all NTT-related attributes.
```
Marvin-Beckmann, line 19: nice, I was wondering whether you thought about this :)

src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// # Panics ...
+    /// - if `modulus` is smaller than `2`.
+    pub fn change_q(&mut self, modulus: impl Into<Modulus>) {
+        self.ntt_basis = Rc::new(None);
+        self.non_zero = Vec::new();
+
+        let new_poly_zq =
+            PolyOverZq::from((self.get_representative_least_nonnegative_residue(), modulus));
+        self.modulus = Rc::new(new_poly_zq);
+    }
```
Marvin-Beckmann, line 48: this would cause problems I think.

The non-zero needs to be set fitting to the modulus, i.e., it captures the non-zero coefficients used for the reduction -> see the reduce call in PolynomialRingZq that uses it. By setting this to 0, I think that essentially the polynomial reduction would always assume that you are using a modulus polynomial of X^N, i.e., there is no proper reduction.

On that note: I just found, that I missed to describe the new attributes from the changed modulus reduction on the Modulus struct.

Please add a test for this case, i.e., do some higher degree modulo multiplication, where the context polynomial is also used for reduction for the tests in the Polynomial constructions

Marvin-Beckmann (changes requested):
I think the current implementation is not working as intended after the mod is changed, see the respective comment

src/integer_mod_q/modulus_polynomial_ring_zq/set.rs
jnsiemer: I've addressed the issue with the `non-zero` vector.
I'd suggest that you address the additionally missing parts described after "On that note" in a separate PR. You have a better idea what's needed / about the edge cases here.

jnsiemer: FYI I changed the `non_zero` vector s.t. it contains all positions of non-zero coefficients in the `ModulusPolynomialRingZq`. Previously, the highest coefficient (always 1) wasn't part of the vector. The tests all still pass, which tells me that either a test reg. this was missing or it didn't change anything and we might be doing double work somewhere now. As the one who wrote this logic, could you please check which one we want / need and add tests accordingly?

Marvin-Beckmann: the leading coefficient after any reduction in my implementation was always set to 0, hence adding to the list the additional non_zero entry that corresponds to the highest degree does not have any impact, as it might change the underlying value first, but after that, it is set to 0 anyways.
Essentially the modification just introduced additional runtime due to double work.

Marvin-Beckmann: Here for the concrete code snippet:
```
    let (poly_len, modulus_polynomial_len) = (poly_len - 1, modulus_polynomial_len - 1);
    if poly_len >= modulus_polynomial_len {
        for i in (modulus_polynomial_len..=poly_len).rev() {
            for k in non_zero {
                unsafe {
                    fmpz_submul(
                        poly_coeffs.add((i - modulus_polynomial_len) + k),
                        poly_coeffs.add(i),
                        modulus_polynomial_coeffs.add(*k),
                    )
                };
            }
            unsafe { fmpz_zero(poly_coeffs.add(i)) };
        }
    }
 ```

